PR #47, opened 2018-11-15
Add CS1591 removal pragma

frblondin:
The goal of this PR is to add pragma instructions that disable & restore [CS1591](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591) warnings. Depending on the project options, missing publicly visible documentation can break the compilation.

Here is an example:
```
partial class ValidatingRecord
#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
{
	public ValidatingRecord(string Name)
	{
		this.Name = Name;
		Validate();
	}

	...

#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
}
```

### DIFF
--- a/src/Amadevus.RecordGenerator.Generators/PartialGeneratorBase.cs
+++ b/src/Amadevus.RecordGenerator.Generators/PartialGeneratorBase.cs
@@ -9,6 +9,9 @@ namespace Amadevus.RecordGenerator.Generators
 {
     internal abstract class PartialGeneratorBase
     {
+        private const string MissingXmlWarning = "CS1591";
+        private const string MissingXmlComment = "// Missing XML comment for publicly visible type or member";
+
         protected PartialGeneratorBase(RecordDescriptor descriptor, CancellationToken cancellationToken)
         {
             Descriptor = descriptor;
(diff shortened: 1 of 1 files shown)


amis92:
That's a good idea. I'd like to hide this behind some kind of a flag, though. I think it could be possible to use an assembly targeted attribute for that, but I haven't explored it deeply yet. Would you mind proposing some solution?

frblondin:
Thanks @amis92 for looking at this PR. This question has been discussed and deeply explored in [here](https://github.com/dotnet/roslyn/issues/12702). There are mainly the following choices:
* Add `GeneratedCodeAttribute` to the type: I don't like this approach as in our case we only want the partial members to be exempted from being documented, not the members being defined in the original partial file created by the developer. The members defined in this partial file should be documented if this documentation check is enabled at the project level
* Add pragmas just like what this PR would do

I'm not sure there are many other options. Why are you reluctant to add this pragma? It has been created for this purpose. Other than having a slightly less readable code I see no drawback here.

amis92:
I think this should actually be done as a part of the framework (`CodeGeneration.Roslyn`). I imagine there would be an assembly attribute that would take `params string[] codes` and add `#pragma warning disable <code>` for all given codes in every generated file. That'd also allow to ignore any other warning in generated code files.

So the solution for your problem would be adding a file like `AssemblyAttributes.cs` with the following contents:
```cs
[assembly: GeneratedCodePragmaDisable("CS1591")]
```

Another approach would be to pass these values as a `.csproj` properties. Not sure which is best.

amis92:
I believe that
* Add Generation of GeneratedCodeAttribute on partial class (#55)
solves the issue this PR attempted to solve as well.

Issue #19, opened 2017-04-18
Add support for IEquatable, override Equals, GetHashCode, ToString


amis92:
All should follow guidelines for records coming in C# 8.0: https://github.com/dotnet/csharplang/blob/master/proposals/records.md

amis92:
It should be opt-out (on by default) as a package, so either generate all, or none. Maybe `SkipEquality` (false by default)?

amis92:
Done, with additional changes and enhancements tracked in #75

PR #55, opened 2019-07-22
Add Generation of GeneratedCodeAttribute on partial class

jflepp:
Add the `GeneratedCodeAttribute` in order to exclude the generated code from code coverage measurements.

### DIFF
--- a/src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
+++ b/src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
@@ -12,32 +12,96 @@ namespace Amadevus.RecordGenerator.Generators
 {
     public static class GeneratedCodeAttributeGeneratorExtensions
     {
-        public static TypeDeclarationSyntax WithGeneratedCodeAttribute(
-            this TypeDeclarationSyntax typeSyntax)
+        private static readonly GeneratedCodeAttributeApplier attributeApplier
+            = new GeneratedCodeAttributeApplier(new GeneratedCodeAttributeGenerator());
+
+        public static ClassDeclarationSyntax WithGeneratedCodeAttributeOnMembers(
+            this ClassDeclarationSyntax typeSyntax)
         {
-            return new GeneratedCodeAttributeGenerator(typeSyntax)
-                .GetWithGeneratedCodeAttribute();
+            return attributeApplier.GetWithGeneratedCodeAttributeOnMembers(typeSyntax);
+        }
+    }
+
+    internal sealed class GeneratedCodeAttributeApplier
+    {
+        private readonly GeneratedCodeAttributeGenerator attributeGenerator;
+
+        private readonly IDictionary<Type, Func<MemberReplacementContext, SyntaxNode>> Updates
+            = new Dictionary<Type, Func<MemberReplacementContext, SyntaxNode>>
+        {
+            {
+                typeof(MethodDeclarationSyntax),
+                (context) => ((MethodDeclarationSyntax)context.Member).WithAttributeLists(context.AttributeGenerator.GenerateAttributeListSyntaxList())
+            },
+            {
+                typeof(PropertyDeclarationSyntax),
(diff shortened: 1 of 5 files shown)


jflepp:
If you like what I did here, I would be interested in implementing the issue #19 as it seems crucial to me.

amis92:
Thanks for the PR. I'm not exactly convinced this should be done this way, though.

My concern is that the attribute applies to all partials of the type, including user-written methods. I understand that for the common case it's natural.

Could you modify this PR to mark all generated *members* (incl. properties and methods, but not the Builder nested class, for example) with that attribute instead of the type itself?

jflepp:
Hi @amis92,

Thank you for your fast response.

Nice catch, I didn't think of that one. I will modify my PR.

Can you tell me why I shouldn't decorate the Builder nested class? As far as I can see that's also just generated code.

Best regards

jflepp:
Hi @amis92,

I updated my PR according to your comment.

best regards

src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
```diff
@@ -12,32 +12,96 @@ namespace Amadevus.RecordGenerator.Generators
...
+        }
 
-        public GeneratedCodeAttributeGenerator(
-            TypeDeclarationSyntax typeSyntax)
+        
+        public ClassDeclarationSyntax GetWithGeneratedCodeAttributeOnMembers(
+            ClassDeclarationSyntax typeDeclaration)
         {
-            this.typeSyntax = typeSyntax;
+            Func<SyntaxNode, bool> toBeReplaced = (node) =>
```
amis92: I feel like those two lambdas would look better as local functions, but it's really a personal preference.

src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
```diff
@@ -12,32 +12,96 @@ namespace Amadevus.RecordGenerator.Generators
...
-                SingletonList(
-                    AttributeList(
-                        SingletonSeparatedList(
-                            Attribute(GenerateQualifiedName())
-                            .WithArgumentList(GenerateAttributeArgumentList())))));
+            return SingletonList(
+                AttributeList(
+                    SingletonSeparatedList(
+                        Attribute(GenerateQualifiedName())
+                            .WithArgumentList(GenerateAttributeArgumentList()))));
```
amis92: The indentation here is off (it need to be decreased for the `WithArgumentsList`.

src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
```diff
@@ -12,32 +12,96 @@ namespace Amadevus.RecordGenerator.Generators
...
     {
-        private readonly TypeDeclarationSyntax typeSyntax;
+        private readonly GeneratedCodeAttributeGenerator attributeGenerator;
+
+        private readonly IDictionary<Type, Func<MemberReplacementContext, SyntaxNode>> Updates
+            = new Dictionary<Type, Func<MemberReplacementContext, SyntaxNode>>
+        {
+            {
+                typeof(MethodDeclarationSyntax),
+                (context) => ((MethodDeclarationSyntax)context.Member).WithAttributeLists(context.AttributeGenerator.GenerateAttributeListSyntaxList())
```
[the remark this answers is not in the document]
jflepp: You're right, I adjusted the code to use the Add method

src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
jflepp: Im used to the lambda style because of the code im usually working in, but don't really care.
I was able to remove one and adjust the other to a local function.

jflepp:
Hi @amis92,

Thank you for your review!
I really appreciate the time you've spend looking at my code.

I hope I've been able to resolve all issues.

Best regards

Issue #45, opened 2018-07-05
Configuration in record attribute


amis92:
A Flags enumeration that by default contains some options.

Also assembly-level attribute support that configures defaults for all records in a given assembly.

amis92:
Regarding parametrizing the `RecordAttribute`, I think the best way forward would be to use a `[Flags]` enum that'll switch on/off various features.

I'd imagine something like that:

```cs
public sealed class RecordAttribute : Attribute
{
    public RecordAttribute()
    {

    }

    public RecordAttribute(Parts parts)
    {
        Parts = parts;
    }

    public Parts Parts { get; } = Parts.Common;
}

[Flags]
public enum Parts
{
    None = 0,
    Constructor = 0b1,
    Withers = 0b10,
    Builder = 0b100,
    Deconstructor = 0b1000,
    Common = Constructor | Withers | Builder | Deconstructor,

    Equals = 0b10000,
    EqualityOperator = 0b100000,
    GetHashCode = 0b1000000,
    Equality = Equals | EqualityOperator | GetHashCode
}
```

What do you think?

_Originally posted by @amis92 in https://github.com/amis92/RecordGenerator/pull/56#issuecomment-514554900_

amis92:
> Regarding parametrizing the `RecordAttribute`, I think the best way forward would be to use a `[Flags]` enum that'll switch on/off various features.

:+1:

> I'd imagine something like that:
>
> ```cs
> [Flags]
> public enum Parts
> {
>     None = 0,
>     Constructor = 0b1,
>     Withers = 0b10,
>     Builder = 0b100,
>     Deconstructor = 0b1000,
>     Common = Constructor | Withers | Builder | Deconstructor,
> 
>     Equals = 0b10000,
>     EqualityOperator = 0b100000,
>     GetHashCode = 0b1000000,
>     Equality = Equals | EqualityOperator | GetHashCode
> }
> ```

I would call it `Features` or `Options` rather than `Parts` because _parts_ is confusing with _partial_. If `Features`/`Options` is too esoteric then `RecordImplementation` would be equally good and explicit.

Nice idea with `Common` but I would just have `Constructor` and `Withers` (which I'd like to suggest renaming to simply `With` although I am all up for creative names) in there unless you would consider adding `Minimal = Constructor | Withers`.

In C#, a `~` member is really called a [_finalizer_][dtor] but [the documentation][dtor] does say that they are also called deconstructors:

[...]

PR #56, opened 2019-07-23
Generate IEquatable.Equals, object.Equals, object.GetHashCode implementations

jflepp:
Implementation of Issue #19 

Work in progress


jflepp:
`Equals` (`IEquotable` & `object`) implemented, not tested yet
`GetHashCode` implemented, not tested yet

The whole branch needs some refactoring.
All implementations used Visual Studio generations as template.

@amis92 
Im not sure wether I should also implement the `ToString()` method. I think at this point we really require some flags.
I am also not sure wether using the `System.Collections.Generics.EqualityComparer` is enough for the `GetHashCode` and `Equals` implementations. I'm working against Visual Studio here because VS sometimes uses operators (I didn't yet figure out when exactly)

amis92:
Ad. `ToString` - Let's leave it out. It's a convenience and it's independent.

---

So, in regards to `EqualityComparer` vs `==` - I think it's mostly to do with whether the type is a well-known type or not.

Honestly, I've spent a few hours today researching equality subject (again!).

My findings:
* [Champion "Support for == and != on tuple types" (15.7)](https://github.com/dotnet/csharplang/issues/190)
* [How to: define value equality for a type (C# Programming Guide)](https://docs.microsoft.com/pl-pl/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type)
* [Champion "Records"](https://github.com/dotnet/csharplang/issues/39)

A random summary of my thoughts:
* Equals is *damn* complicated in a generic I-don't-know-what-it-is case
* For `Equals`, we should probably only use `==` for a couple of well-known types, like integers, bool, string (and their nullables) that are known to have the same behavior of `Equals` and `==` (e.g. `double` doesn't: `double.NaN == double.NaN` is false, while the same with `Equals` is true - I'd expect us to use the `Equals`/`EqualityComparer`).
* For the first iteration we should probably require the class to be `sealed` and not derived in order to have equalities generated. Inheritance is crazy.
* For the `==`/`!=` operators, we could use the ValueTuple compiler trick, e.g. return the result of the `(left.a, left.b, left.c) == (right.a, right.b, right.c)` (see the first link).
* For the `Equals(object)`, `Equals(T)` and `GetHashCode` the current approach is good enough.
  - Optimizing `EqualityComparer` call into `==` could be done later.
  - Introducing support for inheritance could be done later, and even then probably with some restrictions.
* We definitely need a switch/flag to turn this feature on/off/customize.

jflepp:
### Equals implementations performance test

I did a small performance test comparing different `Equals` implementations with a string and an integer. You can lookup the results in [this gist](https://gist.github.com/jflepp/da66bb7a2810a1c7444a845dab51b847).

I interpret the results like this: All implementations take about the same time to complete, with the exception of the `EqualityComparer` which is significantly slower.

My suggestion is that we head for the tuple implementation.

### Inheritance in equals

Implementing `Equals` while respecting inheritance is almost impossible. Nevertheless, I think sealing the class is a too big restriction. Even while I know that I will seal all my classes, that doesn't mean the users will.

IMO we have two ways of finishing up the implementation:

- We leave the implementation as it is. If a class `B` inherits from `A`, it should behave in the same way as `A` does in regard to the inherited functionality. I think of the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) here. So `a.Equals(b)` should return true, because `b` is seen as a implementation of `A` in this case. What is bad, is that `b.Equals(a)` will _likely_ return false which can lead to confusion and unexpected behavior.
- We add a type equality check. So even though `B` inherits from `A` and has the same values on inherited properties, `a.Equals(b)` would fail because `b` is of a different type.

What are your thoughts on this?

atifaziz:
> Implementing `Equals` while respecting inheritance is almost impossible.

I think this says it all; better to not do it than do it wrong.

> Nevertheless, I think sealing the class is a too big restriction. Even while I know that I will seal all my classes, that doesn't mean the users will.

These are _records_ after all, which are tuples with named elements, and as such, it's understandable if they are sealed by default. One can always provide an option to not have them sealed so it's not a restriction (for hacking) but it shouldn't be encouraged by default. I would even go as far as not providing that option initially just in case YAGNI.

From a design perspective (less technical argumentation), a sealed record is not surprising but an open record that doesn't deliver on some expected behaviour across an inheritance hierarchy will lead to disappointment and pain.

> My suggestion is that we head for the tuple implementation.

I think this is a good starting point that can always be improved on.

---
My 2¢ even though you never asked for them. 😉

amis92:
First of all - I agree with @atifaziz. We really shouldn't ignore the possible implications of inheritance on Equals. First and foremost, because if we do it wrong, it'll be inexplicably hard to find what's wrong when someone uses them in dictionaries, or simply List.Contains and it doesn't work.

So, either we generate it by default *but only if* the record is sealed, or we require an opt-in, for which we can raise a warning/error if the class is not sealed.

It's much better to start with a highly restricted feature and widen the possible applications, than produce code which doesn't adhere to .NET's equality rules.

---

And for that reason, tuple-equality-shortcut should be used only for implementations of `==` operator - and not for Equals. The simplest case being a record with a `double` property. Two instances of the same record, which have `double.NaN` as value, must be `Equal` to each other. But, if we provide `==` operator, it should behave similarly to `==` operators of components - and as such, we can use tuple-equality form.

amis92:
Regarding the benchmark, actually the int, string and bool (and all the integer types) together with their nullable counterparts, can always use `==` comparison (and probably should).

All the other types should use `EqualityComparer` or directly implement it's logic. Performance hit is not *that* considerable, although I'd be interested in similar benchmark with more value *and* reference types included.

amis92:
Regarding parametrizing the `RecordAttribute`, I think the best way forward would be to use a `[Flags]` enum that'll switch on/off various features.

I'd imagine something like that:

```cs
public sealed class RecordAttribute : Attribute
{
    public RecordAttribute()
    {

    }

    public RecordAttribute(Parts parts)
    {
        Parts = parts;
    }

    public Parts Parts { get; } = Parts.Common;
}

[Flags]
public enum Parts
{
    None = 0,
    Constructor = 0b1,
    Withers = 0b10,
    Builder = 0b100,
    Deconstructor = 0b1000,
    Common = Constructor | Withers | Builder | Deconstructor,

    Equals = 0b10000,
    EqualityOperator = 0b100000,
    GetHashCode = 0b1000000,
    Equality = Equals | EqualityOperator | GetHashCode
}
```

What do you think?

src/Amadevus.RecordGenerator.Generators/RecordGeneratorOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+                .Parameters.Select((parameter, index) => new
+                {
+                    parameter.Name,
+                    attributeData.ConstructorArguments[index].Value
+                });
+
+            var optionsConstructor = typeof(RecordGeneratorOptions).GetConstructors().Single();
+            var optionsConstructorParameters = optionsConstructor.GetParameters();
+
+            return (RecordGeneratorOptions)optionsConstructor.Invoke(
```
amis92: Nice solution for instantiation! Didn't think of that.

atifaziz:
> Regarding parametrizing the `RecordAttribute`, I think the best way forward would be to use a `[Flags]` enum that'll switch on/off various features.

:+1:

> I'd imagine something like that:
>
> ```cs
> [Flags]
> public enum Parts
> {
>     None = 0,
>     Constructor = 0b1,
>     Withers = 0b10,
>     Builder = 0b100,
>     Deconstructor = 0b1000,
>     Common = Constructor | Withers | Builder | Deconstructor,
> 
>     Equals = 0b10000,
>     EqualityOperator = 0b100000,
>     GetHashCode = 0b1000000,
>     Equality = Equals | EqualityOperator | GetHashCode
> }
> ```

I would call it `Features` or `Options` rather than `Parts` because _parts_ is confusing with _partial_. If `Features`/`Options` is too esoteric then `RecordImplementation` would be equally good and explicit.

Nice idea with `Common` but I would just have `Constructor` and `Withers` (which I'd like to suggest renaming to simply `With` although I am all up for creative names) in there unless you would consider adding `Minimal = Constructor | Withers`.

In C#, a `~` member is really called a [_finalizer_][dtor] but [the documentation][dtor] does say that they are also called deconstructors:

> Finalizers (which are also called **destructors**) are used to perform any necessary final clean-up when a class instance is being collected by the garbage collector.

Consequently, I would rename `Deconstructor` to `Deconstruct`.

[dtor]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/destructors

The equality members (`Equals` and `GetHashCode`) have to work in tandem so I wouldn't give the option to have one implemented and not the other.

I would also recommend inverting some of the flags as an overall design. You want the generator to produce the most sensible and commonly expected features (albeit opinionated?) of a record and opting out some and opting in other nice-to-haves or esoteric cases. In short, I'm thinking this:

```
[Flags]
public enum Features
{
    Basic = 0, // all !excluded + !included
    
    // opt-out
    
    ExcludeConstructor      = 0b0001,
    ExcludeWithers          = 0b0010,
    ExcludeEquality         = 0b0100,
    ExcludeEqualityOperator = 0b1000,
    
    // opt-in

    IncludeBuilder          = 0b0001_0000,
    IncludeDeconstruct      = 0b0010_0000,
    IncludeUpdate           = 0b0100_0000,
}
```

As an aside, shouldn't this design be discussed and iterated in a separate issue as opposed to as part of this PR?

amis92:
Yes, there's actually an issue for that since long time: https://github.com/amis92/RecordGenerator/issues/45

Let's move over that discussion there.

jflepp:
I am not sure wether or not we should leave the equality features as an opt-in. In the end they can make a siginicant difference when working with collections and most people will probably be using the types in collections.

I can agree that the class should be sealed when implementing equality. It is a more restricted approach but it closes a risky area.

amis92:
Let's implement this feature semi-agnostically of that [opt-in/out]. Let's pop a warning into `IProgress` when the class is not sealed (and just skip the generation of Equality), and implement the attribute Flags without any aggregates like Basic/Common. We'll discuss the default settings in #56.

jflepp:
Done some refactorings. More to come.

I would like to write some UnitTests. Can I create a UnitTest project, would you like to create a UnitTest project or would you like to stick to the IntegrationTests?

A quick summary of the things discussed above regarding the equality implementation in this first iteration:

- We will implement operator equality operations for well-known types.
- We will stick to the `EqualityComparer` for all other types.
- We will seal the class if equality operations are being used by the user.
- We will not merge this PR before flags have been implemented.

Please correct me if I misunderstood something.

amis92:
> I would like to write some UnitTests. Can I create a UnitTest project, would you like to create a UnitTest project or would you like to stick to the IntegrationTests?

I'll create a UnitTest project.
 
> We will implement operator equality operations for well-known types.

We'll *use* `==` operators for welll known types. I guess that's what you meant.

> We will stick to the EqualityComparer for all other types.

Yes. Although I've been reading though https://github.com/dotnet/csharplang/blob/master/proposals/records.md and below the [record struct example](https://github.com/dotnet/csharplang/blob/master/proposals/records.md#record-struct-example) there's another implementation that makes sense:

`return other != null && Equals(Id, other.Id) && Equals(Name, other.Name)`

This uses [`object.Equals(object, object)` method](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=netframework-4.8#System_Object_Equals_System_Object_System_Object_)

I don't know if it'll be faster than EqualityComparer, and this'll box structs could be used only for reference classes. Just a thought.

> We will seal the class if equality operations are being used by the user.

I don't agree, we shouldn't seal classes ourselves when needed. Rather, if the customer requests equality (maybe by default), on a non-sealed class, we'll warn that the given class must be sealed for equality to be generated.

> We will not merge this PR before flags have been implemented.

Probably. I'll do the unit tests and flags today or tomorrow.

src/Amadevus.RecordGenerator.Generators/Names.cs
```diff
@@ -17,5 +17,8 @@ internal static class Names
 
         public const string ToolName = "Amadevus.RecordGenerator";
         public const string GeneratedCodeAttribute = "System.CodeDom.Compiler.GeneratedCodeAttribute";
+
+        public const string EqualityCompararerQualifiedName = "System.Collections.Generic.EqualityComparer";
+        public const string IEquotableQualifiedName = "System.IEquatable";
```
amis92: `IEquatable`, typo

src/Amadevus.RecordGenerator.Generators/Names.cs
```diff
@@ -17,5 +17,8 @@ internal static class Names
 
         public const string ToolName = "Amadevus.RecordGenerator";
         public const string GeneratedCodeAttribute = "System.CodeDom.Compiler.GeneratedCodeAttribute";
+
+        public const string EqualityCompararerQualifiedName = "System.Collections.Generic.EqualityComparer";
```
amis92: `Comparer`, typo

src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
```diff
@@ -119,21 +119,9 @@ private AttributeArgumentListSyntax GenerateAttributeArgumentList()
...
-                namespaceAndIdentifier.First(),
-                namespaceAndIdentifier.Skip(1).First());
-
-            foreach (var identifier in namespaceAndIdentifier.Skip(2))
-            {
-                attributeName = QualifiedName(attributeName, identifier);
-            }
-
-            return attributeName; 
+            return QualifiedNameGenerator.GenerateQualifiedName(Names.GeneratedCodeAttribute);
```
amis92: I think you can just do [`ParseName`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntaxfactory.parsename?view=roslyn-dotnet) instead of creating a whole custom class.

src/Amadevus.RecordGenerator.Generators/SyntaxSnippetGenerators.cs
```diff
@@ -0,0 +1,155 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Amadevus.RecordGenerator.Generators
+{
```
amis92: I don't like those helper classes. Soon we'll be able to have dependencies in generators and I'd then use SyntaxGenerator probably, but until then let's keep `SyntaxExtensions` for API touchups and not build those big helpers.

src/Amadevus.RecordGenerator.Generators/SyntaxExtensions.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
```
amis92: Not sure why this appeared here… No other changes in this file?

src/Amadevus.RecordGenerator.Generators/Names.cs
jflepp: Fixed

src/Amadevus.RecordGenerator.Generators/Names.cs
jflepp: Fixed

src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
jflepp: Didn't know of that one. Used ist whereever possible

src/Amadevus.RecordGenerator.Generators/SyntaxSnippetGenerators.cs
jflepp: Removed helper classes, merged equality generation into single class

src/Amadevus.RecordGenerator.Generators/SyntaxExtensions.cs
jflepp: Removed and sorted usings in all projects

jflepp:
Note: this PR is still a work in progress. I will do more refactorings, write tests, increase functionality etc.

Currently I'm implementing the sealed analyzer, altough I won't be able to finish it before options are available.

best regards

amis92:
Sealed-analysis can be done not as an Analyzer per se, but rather as a Warning Diagnostic returned via IProgress passed to the GenerateAsync method in the main RecordGenerator. We just need to define the DiagnosticDescriptor and if Equality is requested to be generated, we send the warning into the `progress` instance. Should work.

amis92:
And to keep your motivation up, I'm really glad you took up this challenge. It's a complicated feature with a lot of pesky details to work out. I'm very happy we have this discussion and you're still on track to break that wall! 🚀

jflepp:
Thank you for your support :smile:

Im currently on a kind of vacation, so I got enough time to spend here. I also want to bring in the functionality I find necessary in order to discuss this package with my team at work.

At the moment the task seems doable to me, so I'm surely gonna finish this.

I also appreciate the time you're spending reviewing my progress and advising me, so thanks for that 👍

jflepp:
@amis92 
What do you think of this shared source file approach (in [this commit](https://github.com/amis92/RecordGenerator/pull/56/commits/2edb935f3b1eb83f6e7b8d1b956a8f663a388756))?
I wanted to define all diagnostic descriptors we provide in a single location.

best regards

amis92:
Not sure. I think it's ok to have the file shared, but I'd prefer for the file to still be directly in Analyzers project and from there linked into Generators. So the Generators would just "borrow" them, but they'd actually still belong to Analyzers.

src/Amadevus.RecordGenerator.Generators/EqualityUsageAnalyzer.cs
```diff
@@ -0,0 +1,27 @@
...
+    internal sealed class EqualityUsageAnalyzer
+    {
+        public static IEnumerable<Diagnostic> GenerateDiagnostics(RecordDescriptor descriptor)
+        {
+            var sealedToken = SyntaxFactory.Token(SyntaxKind.SealedKeyword);
+            if (descriptor.TypeDeclaration.Modifiers.Contains(sealedToken)) yield break;
+
+            yield return Diagnostic.Create(
+                Descriptors.X1001_RecordMustBeSealedIfEqualityIsEnabled,
+                descriptor.TypeDeclarationWithTrivia.Identifier.GetLocation(),
```
amis92: Why is *WithTrivia necessary?

jflepp: Because I needed to call `GetLocation` on the original identifier, otherwise the warning would appear somewhere in the using declarations.
If I created a property `Location` and called `GetLocation` while creating the Descriptor, there is a chance that the Location isn't even needed and would therefore be calculated unnecessarily.
But it's your decision how we're going to implement it.

jflepp:
@amis92,
Could you take a look at my implementation of the equals operator usage?
Maybe I missed something, but I couldn't figure out how one would compare `SyntaxTokens` of predefined types.
In the end I implemented it using string comparisons...

thanks and best regards

jflepp:
Update: I found out that the Semantic Model is available to us, and this is exactly what I need.
I will use the semantic model to identify the well-known types.

src/Amadevus.RecordGenerator.Generators/EqualityUsageAnalyzer.cs
amis92: Right. That makes sense.

With that in mind, please rename those things to something more self-descriptive. `SourceTypeDeclaration` or `OriginalTypeDeclaration`, something like that.

src/Amadevus.RecordGenerator.Generators/EqualityUsageAnalyzer.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading.Tasks;
+
+namespace Amadevus.RecordGenerator.Generators
+{
+    internal sealed class EqualityUsageAnalyzer
+    {
+        public static IEnumerable<Diagnostic> GenerateDiagnostics(RecordDescriptor descriptor)
+        {
+            var sealedToken = SyntaxFactory.Token(SyntaxKind.SealedKeyword);
+            if (descriptor.TypeDeclaration.Modifiers.Contains(sealedToken)) yield break;
```
amis92: This doesn't work. Try:

```csharp
            if (descriptor.TypeDeclaration.Modifiers.Any(token => token.IsKind(SyntaxKind.SealedKeyword))) yield break;
```

`Contains` uses default equality, and it seems those tokens are not equal almost all the time (maybe whitespace trivia is the reason). Anyway, my snippet solves the problem.

Not sure how, but this could probably do with some tests...

jflepp: You were right, I implemented it the way you suggested.

Yes, the whole feature sill needs to be tested. Tests will come as soon

jflepp:
- Added Semantic Analysis to Descriptor.
- Implemented well-known type recognition based on qualified name.
- Using operator equality comparison for well-known types
- Fixed EqualityUsageAnalyzer according to your suggestion

Not sure why the build breaks

amis92:
Because there are merge conflicts. 
image

src/Amadevus.RecordGenerator.Generators/SyntaxExtensions.cs
```diff
@@ -148,6 +148,13 @@ public static NameSyntax GetTypeSyntax(this TypeDeclarationSyntax typeDeclaratio
             return SyntaxFactory.GenericName(identifier, typeArgList);
         }
 
+        public static string GetQualifiedName(this ISymbol symbol)
+        {
+            var symbolDisplayFormat = new SymbolDisplayFormat(
+                typeQualificationStyle: SymbolDisplayTypeQualificationStyle.NameAndContainingTypesAndNamespaces);
+            return symbol.ToDisplayString(symbolDisplayFormat);
```
amis92: This needs to be tested with nullables in both full-format and sugar-format (`Nullable<int>` and `int?`).

jflepp: This I don't expect it to make a difference, because we're working agains semantic data here (sugar shouldn't matter)
But I will surely test it 👍

jflepp: This I don't expect it to make a difference, because we're working agains semantic data here (sugar shouldn't matter)
But I will surely test it 👍

amis92: The reason I ask is because I don't really know the behavior of those options. Since the `generic` option by default says "none", will `int?` be printed as `System.Nullable`, `System.Nullable<int>`, `int?` (don't think so), or ``System.Nullable`1``?

amis92: There are also some readily available formats on the `SymbolDisplayFormat` type, maybe one of those would be enough? Otherwise, this is something worth making a static field for.

jflepp:
Yes there are some more readable methods, but actually no they do not return the full qualified name for internal types. 
A static field sounds like a good idea. 

I will look into this later, but as far as I've seen it always return the name with a ? as suffix. The symbol should no be coupled to the syntax

src/Amadevus.RecordGenerator.Generators/SyntaxExtensions.cs
jflepp: Yes there are some more readable methods, but actually no they do not return the full qualified name for internal types.
A static field sounds like a good idea.

I will look into this later, but as far as I've seen it always return the name with a ? as suffix. The symbol should no be coupled to the syntax

amis92:
With Unit Tests project in place, as well as Features flags implemented, I think this draft can be now published as actual PR, pending rebase/merge with current master :)

Let's finish this! 🚀 :shipit:

jflepp:
Merged the latest changes into the fork. I will be implementing the features according to the new feature flags soon 👍

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -1,7 +1,6 @@
 ﻿using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using System;
```
amis92: This change is out-of-context. Please revert.

src/Amadevus.RecordGenerator.Generators/PartialGeneratorBase.cs
```diff
@@ -1,9 +1,6 @@
 ﻿using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using System;
-using System.Linq;
-using System.Reflection;
```
amis92: This change is out-of-context. Please revert.

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -49,9 +49,9 @@ public enum Features
...
         Default = Constructor | Withers | ToString | Builder | Deconstruct,
 
-        //ObjectEquals = 0b_100000,
-        //EquatableEquals = 0b_1000000,
-        //OperatorEquals = 0b_10000000,
-        //Equality = ObjectEquals | EquatableEquals | OperatorEquals
+        ObjectEquals = 0b_100000,
+        EquatableEquals = 0b_1000000,
+        OperatorEquals = 0b_10000000,
+        Equality = ObjectEquals | EquatableEquals | OperatorEquals
```
amis92: Just a note to add xml-doc comments. Also, I think with the longer numbers, we can group into 4-digit groups separated by `_`. When you're at it, could you fix Deconstruct as well? :) Thanks!

src/Amadevus.RecordGenerator.Generators/BuilderPartialGenerator.cs
```diff
@@ -1,7 +1,6 @@
 ﻿using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using System;
```
amis92: This change is out-of-context. Please revert.

src/Amadevus.RecordGenerator.Generators/GeneratedCodeAttributeGenerator.cs
```diff
@@ -119,21 +118,9 @@ private AttributeArgumentListSyntax GenerateAttributeArgumentList()
...
-            foreach (var identifier in namespaceAndIdentifier.Skip(2))
-            {
-                attributeName = QualifiedName(attributeName, identifier);
-            }
-
-            return attributeName; 
+            return ParseName(Names.GeneratedCodeAttribute);
         }
 
         private string GetToolVersion() => GetType().GetTypeInfo().Assembly.GetName().Version.ToString();
```
amis92: While refactoring the nearby code, this could be simplified into `ThisAssembly.AssemblyVersion` generated by Nerdbank.GitVersioning, and also then since it's constant, probably inlined completely at usage place.

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
jflepp: I'd like to refer to the boy scout rule here: [Clean Code, Robert C. Martin](http://www.informit.com/articles/article.aspx?p=1235624&seqNum=6&source=post_page). It doesn't belong in this Pull-Request but does a small clean-up, so why not take it?

src/Amadevus.RecordGenerator.Generators/PartialGeneratorBase.cs
jflepp: See comment above.

src/Amadevus.RecordGenerator.Generators/BuilderPartialGenerator.cs
jflepp: see comment above

src/Amadevus.RecordGenerator.Attributes/Features.cs
jflepp: 👍

jflepp:
I wrote some XML-style comments.
I know that I tend to write too large comments, because I don't see a value in XML comments if they don't _really_ provide more information than the name already does.
Feel free to shorten it if you wish to

jflepp:
Note: The implementations of the EquatableEquals and OperatorEquals will depend on the ObjectEquals.
Should I change this behaviour (which would lead to more code) or should I write an analyzer which checks the Features?

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,45 @@ public enum Features
...
         /// Default feature set.
         /// </summary>
         Default = Constructor | Withers | ToString | Builder | Deconstruct,
 
-        ObjectEquals = 0b_100000,
-        EquatableEquals = 0b_1000000,
-        OperatorEquals = 0b_10000000,
+        /// <summary>
+        /// <see cref="object.Equals(object)"/> and <see cref="object.GetHashCode()"/> overrides. 
+        /// <see cref="object.Equals(object)"/> implements a equality comparison which compares
```
amis92: nit: *an* equality comparison

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,45 @@ public enum Features
...
+
+        /// <summary>
+        /// <see cref="IEquatable{T}.Equals(T)"/> implementation that provides a call to the 
+        /// <see cref="object.Equals(object)"/> method in a type safe way. 
+        /// Is dependent on the feature <see cref="ObjectEquals"/>.
+        /// </summary>
+        EquatableEquals = 0b_1000_000,
+
+        /// <summary>
+        /// Overrides a the <see langword="==" /> by using the <see cref="object.Equals(object)"/> method.
```
amis92: nit: drop *a* or *the*

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,45 @@ public enum Features
...
-        ObjectEquals = 0b_100000,
-        EquatableEquals = 0b_1000000,
-        OperatorEquals = 0b_10000000,
+        /// <summary>
+        /// <see cref="object.Equals(object)"/> and <see cref="object.GetHashCode()"/> overrides. 
+        /// <see cref="object.Equals(object)"/> implements a equality comparison which compares 
+        /// all record entries (properties that are read-only and auto-implemented) 
+        /// of the other <see cref="object"/> with it's own values. 
+        /// <see cref="object.GetHashCode()"/> implements a hash code calculation which calculates
+        /// a hash code using all record entries.
```
amis92: >implements a hash code calculation which calculates a hash code

I'd simplify to

> `object.GetHashCode()` override which calculates a hash code using all record properties.

amis92: Similar simplification for other comments could be done.

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
amis92: This was mostly because of #71/#74 which will probably go into conflict with these changes. Up to you. Resolve the comments however you see fit.

amis92:
Thanks, looking good.

src/Amadevus.RecordGenerator.Generators/EqualityUsageAnalyzer.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            if (!descriptor.Features.HasFlag(Features.EquatableEquals) 
+             && !descriptor.Features.HasFlag(Features.ObjectEquals)
+             && !descriptor.Features.HasFlag(Features.OperatorEquals)
+             || descriptor.TypeDeclaration.Modifiers.Any(t => t.IsKind(SyntaxKind.SealedKeyword)))
+            {
+                yield break;
+            }
+
+            yield return Diagnostic.Create(
```
amis92: Lets put this `yield return` inside the `if`.

Then we can add more conditions with more possible diagnostics.

And the next diagnostic needed is that the class cannot have a base class (other than default `object`). I've done it here:

https://github.com/WarHub/wham/blob/af6b15ec1ee8c988e59b4f4223e3bb8673d8a06d/src/WarHub.ArmouryModel.Source.CodeGeneration/Generators/PartialGeneratorBase.cs#L18

> `IsDerived = BaseType.SpecialType != SpecialType.System_Object;`

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -0,0 +1,231 @@
...
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Amadevus.RecordGenerator.Generators
+{
+    internal abstract class EqualityPartialGeneratorBase : PartialGeneratorBase
+    {
+        protected const string EqualsMethodName = "Equals";
+        protected const string GetHashCodeMethodName = "GetHashCode";
+
+        protected string[] wellKnownTypes { get; } = new[] { typeof(bool), typeof(byte), typeof(sbyte), typeof(char), typeof(int), typeof(uint), typeof(long), typeof(ulong), typeof(short), typeof(ushort), typeof(string) }
```
amis92: A property should be CamelCased. Also I'd prefer `ImmutableArray` here because in array you can change the value by accident.
Other thing: the name could be more useful. Currently it's not correct, because there are many more well known types (non-integral) - so I'd suggest something closer to `OperatorEqualityTypes` that's associated with operator equality used on them.

amis92:
We need to finish off this PR ASAP, I feel it's getting too long.

Let's touch up cosmetics, test it manually and merge it. Then we can work on fixing up the smaller stuff, because the longer I browse through the more issues I see. We'll also be able to parallelize the work on those additional issues and/or writing tests. Can you complete this PR soon?

jflepp:
Hi @amis92,

I agree with you that the PR should be finished soon. 

Unfortunately there are, even when ignoring refactorings and tests, 2 important changes left: 
- Generate standalone ObjectEquals if EquatableEquals is not selected
- Generate operator equality

I think these two points must certainly be in a working state before releasing.

Unfortunately, today is the swiss national day so I'll be busy and I'm flying into vacation next week. 

I will put the PR out of Draft, so everybody is welcome to help in order to finish this asap. 

Another option would be to wait another 2 weeks before releasing, until then I'd be able to finish everything in a certain quality. 

best regards

jflepp:
Implemented standalone `Equals(object)` generation

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
jflepp: Actually, Microsoft conventions do not specify wether private / protected members should be uppercase or lowercase. In my company we always use lowercase names for internal members, but I don't really care as long as it's used consistently.

jflepp:
Implemented operator equals

jflepp:
The operator Equals implementation uses (as VS does) the `EqualityComparer` which relies on the `Equals` implementations of the objects. Thus, it is recommended to implement `Equals` when using equality operators.

Fortunately, Roslyn already throws a warning CS0661 so we don't have to.

jflepp:
@amis92, 
It's a bit of a mess, but if you want to push for this feature I think it's in a state ready for reviewing / manually testing now

jflepp:
How do you feel about creating more interfaces?

I get the feeling that I'm not able to create clean UnitTests because everything is coupled so tightly.

A starting point could be a `IRecordDescriptor` interface. This way we could created easier, cleaner UnitTests.

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,44 @@ public enum Features
...
         Default = Constructor | Withers | ToString | Builder | Deconstruct,
 
-        //ObjectEquals = 0b_100000,
-        //EquatableEquals = 0b_1000000,
-        //OperatorEquals = 0b_10000000,
-        //Equality = ObjectEquals | EquatableEquals | OperatorEquals
+        /// <summary>
+        /// <see cref="object.Equals(object)"/> and <see cref="object.GetHashCode()"/> overrides. 
+        /// <see cref="object.Equals(object)"/> implements an equality comparison which compares 
+        /// all record entries properties.
```
amis92: Weird line break

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,44 @@ public enum Features
         /// <c>Deconstruct</c> method:
         /// <see href="https://docs.microsoft.com/pl-pl/dotnet/csharp/deconstruct#deconstructing-user-defined-types"/>
         /// </summary>
-        Deconstruct = 0b_10000,
+        Deconstruct = 0b_1000_0,
```
amis92: Not the formatting I thought about. I'd expect left-aligned breaks, e.g.
* `0b_1_000`
* `0b_10_000`
* `0b_100_000`
* `0b_1000_000`

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,44 @@ public enum Features
...
         /// </summary>
         Default = Constructor | Withers | ToString | Builder | Deconstruct,
 
-        //ObjectEquals = 0b_100000,
-        //EquatableEquals = 0b_1000000,
-        //OperatorEquals = 0b_10000000,
-        //Equality = ObjectEquals | EquatableEquals | OperatorEquals
+        /// <summary>
+        /// <see cref="object.Equals(object)"/> and <see cref="object.GetHashCode()"/> overrides. 
+        /// <see cref="object.Equals(object)"/> implements an equality comparison which compares
```
amis92: *~implements an equality comparison which~ compares […]*

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,44 @@ public enum Features
...
-        //EquatableEquals = 0b_1000000,
-        //OperatorEquals = 0b_10000000,
-        //Equality = ObjectEquals | EquatableEquals | OperatorEquals
+        /// <summary>
+        /// <see cref="object.Equals(object)"/> and <see cref="object.GetHashCode()"/> overrides. 
+        /// <see cref="object.Equals(object)"/> implements an equality comparison which compares 
+        /// all record entries properties.
+        /// of the other <see cref="object"/> with it's own values. 
+        /// <see cref="object.GetHashCode()"/> override which calculates a hash code using all record properties.
+        /// <see cref="object.GetHashCode()"/>
```
amis92: Unnecessary duplicate?

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,44 @@ public enum Features
...
+        /// </summary>
+        /// <remarks>
+        /// <see cref="object.Equals(object)"/> returns <see langword="true" /> if all values match 
+        /// and returns <see langword="false"/> if <see langword="null" /> is passed, another type is passed or at least one value doesn't mach.
```
amis92: End of sentence: *ma**t**ch*

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,44 @@ public enum Features
...
+        /// </remarks>
+        ObjectEquals = 0b_1000_00,
+
+        /// <summary>
+        /// <see cref="IEquatable{T}.Equals(T)"/> implementation that provides a call to the 
+        /// <see cref="object.Equals(object)"/> method in a type safe way. 
+        /// Is dependent on the feature <see cref="ObjectEquals"/>.
```
amis92: Shouldn't it actually be the other way around? I thought `ObjectEquals` would call into `EqualityEquals`. To be thought about.

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -42,16 +42,44 @@ public enum Features
...
+        /// </summary>
+        EquatableEquals = 0b_1000_000,
+
+        /// <summary>
+        /// Overrides the <see langword="==" /> operator by using the <see cref="object.Equals(object)"/> method.
+        /// Is dependent on the feature <see cref="ObjectEquals"/>.
+        /// </summary>
+        OperatorEquals = 0b_1000_0000,
```
amis92: I thought we agreed to use direct operator equality/valuetuple equality. That might come later.

src/Amadevus.RecordGenerator.Generators/EqualityUsageAnalyzer.cs
```diff
@@ -0,0 +1,28 @@
...
+                Descriptors.X1001_RecordMustBeSealedIfEqualityIsEnabled,
+                descriptor.TypeDeclarationLocation,
+                descriptor.TypeIdentifier.Text);
+        }
+        
+    }
```
amis92: In the end, I think we may want to keep analysis in the Analyzers project anyway. There we can offer a CodeFix that'll make the class sealed, enhancing the experience.

src/Amadevus.RecordGenerator.Attributes/Features.cs
amis92: Moved to #75

src/Amadevus.RecordGenerator.Generators/EqualityUsageAnalyzer.cs
amis92: Moved to #76

PR #58, opened 2019-07-24
Generate ToString() for records

atifaziz:
This PR adds to #19 by implementing an override for `ToString()`.

The implementation I've added is a quick cheat (that can improved upon) that piggy backs on the built-in implementation of `ToString()` generated by compiler for anonymous types, which is the closest thing we had to records in C# so it seems reasonable to borrow from there. So, say, the generated implementation for the test `Item` will look as follows:

```c#
public override string ToString() => new { Id, Name }.ToString();
```

This gives you:

    { Id = item1, Name = item name }

For `Container`, however, the formatted string becomes:

    { Id = cont1, Name = container name, Items = System.Collections.Immutable.ImmutableArray`1[System.Int32] }

This is not great, but not any different than what you get from anonymous records/types.

The one downside of this implementation is the extra object allocation for the anonymous object but I am hoping that's something one can improve on after feature completion.

### DIFF
--- a/test/Amadevus.RecordGenerator.Test/RecordFormatterTests.cs
+++ b/test/Amadevus.RecordGenerator.Test/RecordFormatterTests.cs
@@ -3,6 +3,9 @@ using Xunit;
 
 namespace Amadevus.RecordGenerator.Test
 {
+    using System.Collections.Generic;
+    using TestsBase;
+
     public class RecordFormatterTests : RecordTestsBase
     {
         [Fact]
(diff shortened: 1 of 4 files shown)


test/Amadevus.RecordGenerator.Test/RecordFormatterTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            container.WithItems(items);
+            var shape = new
+            {
+                Id = ContainerId,
+                Name = ContainerName,
+                Items = items.ToString(),
+            };
+            Assert.Equal(shape.ToString(), container.ToString());
+        }
+    }
```
amis92: Could you add a test that uses GenericRecord with Item type parameter to validate nested record case?

atifaziz: Done in 4245fa13a721d5b7634abde68c60ca1cfb44a9c3.

amis92:
One last thing, could you add a note to the CHANGELOG.md file in Unreleased section that this feature was added? Thanks!

atifaziz:
> One last thing, could you add a note to the CHANGELOG.md file in Unreleased section that this feature was added? Thanks!

Done with c2ee802289861c96350be31accacf38d60997f54.

Issue #57, opened 2019-07-24
Use camelCase instead of PascalCase parameter names


atifaziz:
The naming convention in .NET for parameters is to use camel case (`fooBar`) but the generator uses Pascal case (`FooBar`) like properties do.

I'd be happy to submit a PR to fix this if it would be welcome.

amis92:
Well, yeah. A PR will be welcome, definitely. I'd like to keep it super-simple, so lower-casing the first character.


Also, this PR will need as a pre-requisite the #28. Can you do that as well?

atifaziz:
> Also, this PR will need as a pre-requisite the #28. Can you do that as well?

I'm afraid I am not familiar with analyzers so it might take me more time. I can try but otherwise should this stall until #28 is done? Or can I still submit a PR for this and you merge once #28 is done? Thinking how we can avoid a waterfall approach here.

amis92:
Sure, go on with this, and I'll add the analyzer to your PR maybe, so remember to check "allow contributor to make changes" or something when opening PR.

amis92:
Although feel free to try the Analyzers yourself, it's not so hard and quite fun :D

atifaziz:
> Although feel free to try the Analyzers yourself, it's not so hard and quite fun :D

Sounds like a great opporunity to do this together using [VS Live Share](https://visualstudio.microsoft.com/services/live-share/) if it's not hard and it would certainly be a fun experience. 😉

amis92:
That'd be great! Do you use gitter or other platform with PMs so we can agree on timeframe and other details?

atifaziz:
> That'd be great!

Awesome!

> Do you use gitter…

Yes, I do!

Issue #28, opened 2018-01-25
Add analyzer checking property names are all case-ignore different


amis92:
Record Properties must not differ by case only.

PR #59, opened 2019-07-24
Use the more conventional camel case for generated parameters

atifaziz:
This PR fixes issue #57.


src/Amadevus.RecordGenerator.Generators/CSharpKeyword.cs
```diff
@@ -0,0 +1,99 @@
...
+    static class CSharpKeyword
+    {
+        // The following method was derived from:
+        // Source https://github.com/dotnet/roslyn/blob/9d80dea7fe1b14043b9b2ac4d0b59ed26f508742/src/Tools/Source/CompilerGeneratorTools/Source/BoundTreeGenerator/BoundNodeClassWriter.cs#L1821-L1905
+        //
+        // Copyright (c) Microsoft. All Rights Reserved.
+        // Licensed under the Apache License, Version 2.0:
+        // https://www.apache.org/licenses/LICENSE-2.0
+
+        public static bool Is(this string name)
```
amis92: I really don't like this approach, but I'm not sure there's any better way.

src/Amadevus.RecordGenerator.Generators/Lazy.cs
```diff
@@ -0,0 +1,30 @@
...
+            Func<TArg, TResult> factory)
+        {
+            if (factory == null) throw new ArgumentNullException(nameof(factory));
+
+            var (initialized, value) = target;
+            if (!initialized)
+                target = (true, value = factory(arg));
+            return value;
+        }
+    }
```
amis92: Do we need our own `Lazy`? Wy not use the framework one?

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
```diff
@@ -33,6 +36,14 @@ public Entry(SyntaxToken Identifier, TypeSyntax Type, PropertyDeclarationSyntax
 
             public SyntaxToken Identifier { get; }
 
+            public SyntaxToken IdentifierInCamelCase =>
+                Lazy.EnsureInitialized(ref identifierInCamelCase, this, self =>
+                {
+                    var id = (string)self.Identifier.Value;
+                    var id2 = char.ToLowerInvariant(id[0]) + id.Substring(1);
+                    return SyntaxFactory.Identifier(CSharpKeyword.Is(id2) ? "@" + id2 : id2);
+                });
```
amis92: It's really not a very costly method. I think we can initialize this property directly in a constructor.

src/Amadevus.RecordGenerator.Generators/CSharpKeyword.cs
amis92: So, I approached this problem already with https://github.com/amis92/RecordGenerator/commit/1172c747b1b1a5fd669b86299eca53ee21f5fc05 (using MS.CA.CSharp.Workspaces package). I'm considering keeping this PR open because I might be able to allow dependencies with CG.R soon, and then it'll be as nice as in the referenced commit.

src/Amadevus.RecordGenerator.Generators/Lazy.cs
atifaziz: This one avoids allocations and is much simpler to read, use and reason about. Since `SyntaxNode` is a `struct`, you need the following [overload of `Lazy.EnsureInitialized`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazyinitializer.ensureinitialized?view=netframework-4.8#System_Threading_LazyInitializer_EnsureInitialized__1___0__System_Boolean__System_Object__System_Func___0__) if you want to go with what's provided by the system libraries:

```c#
public static T EnsureInitialized<T> (ref T target,
                                      ref bool initialized,
                                      ref object syncLock,
                                      Func<T> valueFactory);
```

This requires you to introduce 3 _separately named_ fields in the object, which is ugly! It also forces a lock when it's not needed most of the time. The allocation (a minor pick, really) comes from `valueFactory` inducing a closure in the most practical cases.

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
atifaziz: I'll leave that decision to you and adjust accordingly. Just let me know what you wish to proceed with finally. Also, sorry about the `id2`, which was a horrible slip.

src/Amadevus.RecordGenerator.Generators/CSharpKeyword.cs
atifaziz: Ideally, Roslyn would have such a method. What your biggest concern? There have been no new keywords in C# for years; most if not all have been introduced as contextual keywords. It's not like this list needs maintaining a lot and it can be removed the day Roslyn has such an API. Meanwhile, happy to consider another approach you have to suggest.

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
atifaziz: > Also, sorry about the `id2`, which was a horrible slip.

FWIW, fixed in dd117f64a1fa3fd666ce33a888364dd5c983cdd2.

src/Amadevus.RecordGenerator.Generators/Lazy.cs
amis92: It's an interesting observation. Still, I think the single use case can be un-lazied, inlined into Entry constructor, and we can forget about anything Lazy. :)

src/Amadevus.RecordGenerator.Generators/CSharpKeyword.cs
```diff
@@ -0,0 +1,99 @@
...
+                case "using":
+                case "class":
+                case "struct":
+                case "interface":
+                case "delegate":
+                case "checked":
+                case "get":
+                case "set":
+                case "add":
+                case "remove":
```
amis92: `get, set, add, remove` should be dropped (or maybe commented out), they're legal for parameter name.

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
amis92: Let's do initialization in the constructor, the implementation can be moved to a static method.

amis92: Another reason is that this is used in the generated constructor, so in 99% of scenarios it'll get used anyway.

src/Amadevus.RecordGenerator.Generators/CSharpKeyword.cs
amis92: Let's leave it as it is for time being. I've created an issue for future enhancement in this area: #62

src/Amadevus.RecordGenerator.Generators/Lazy.cs
atifaziz: > Still, I think the single use case can be un-lazied, inlined into Entry constructor, and we can forget about anything Lazy. :)

Sure, I don't want to over-engineer for the sake of it.

atifaziz: Revert to eager initialization with 2f13bbd5f7d91cccdd499aa083a93d1da10bdf54.

src/Amadevus.RecordGenerator.Generators/CSharpKeyword.cs
atifaziz: Good catch! Indeed they're [contextual](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/#contextual-keywords) and disappointed that borrowing from Roslyn code base still requires review!

Fixed with 6ac1efb5c03ab3404fed4f02139a6cdb3c19ca02.

atifaziz:
> Also, please add Changelog entry :)

Done with 5506650b338b07abfc8249b39a973e13a2034a55.

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
atifaziz: Yeah, except I was thinking of those cases where people would be choosing `None` for `Features` 😜; apologies for the poor taste in or tongue-in-cheek humour.

test/Amadevus.RecordGenerator.Test/RecordConstructorTests.cs
```diff
@@ -9,7 +9,7 @@ public class RecordConstructorTests : RecordTestsBase
         [Fact]
         public void Ctor_HasParametersNamedLikeProperties()
         {
-            new Item(Id: ItemId, Name: ItemName);
+            new Item(id: ItemId, name: ItemName);
```
amis92: While at it, could you add tests that check camelCase names for all generated methods types as well?

atifaziz: I've added tests and fixed the issue in 8ac523cfd654538a700d8e4d987470746f127dcb by using `value` as the parameter name for the all `With*` methods. I don't find there's any value in naming them uniquely based on the property name.

test/Amadevus.RecordGenerator.Test/RecordNamingConventionsTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                typeof(Item)           , typeof(Item.Builder),
+                typeof(Container)      , typeof(Container.Builder),
+                typeof(GenericRecord<>), typeof(GenericRecord<>.Builder),
+            }
+            from m in t.GetMethods(BindingFlags.DeclaredOnly | BindingFlags.Public
+                                                             | BindingFlags.Instance
+                                                             | BindingFlags.Static)
+            from p in m.GetParameters()
+            select new[] { t.Name, m.Name, p.Name };
```
amis92: Love that approach!

src/Amadevus.RecordGenerator.Generators/Names.cs
```diff
@@ -1,5 +1,8 @@
 ﻿namespace Amadevus.RecordGenerator.Generators
 {
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
```
amis92: Move usings outside namespace here as well

amis92: But also, remove them as unused, actually. :D

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -112,7 +112,7 @@ MethodDeclarationSyntax CreateRecordMutator(RecordDescriptor.Entry entry)
                     .AddModifiers(SyntaxKind.PublicKeyword)
                     .WithParameters(
                         Parameter(
-                            entry.Identifier)
+                            Identifier(Names.Value))
```
amis92: That's an important change. Could you note it additionally down in changelog?

Also, just from diff, shouldn't you update `arguments` creation as well? The tests don't pass.

Which also triggered me to check why the AppVeyor succeeded and it turns our AppVeyor discovers 0 tests... Off to fixing that.

amis92: I think I lost some words on the way from thought to typing: the `value` is not passed into Update call, which results in the current property value being passed instead.

src/Amadevus.RecordGenerator.Generators/Names.cs
atifaziz: > But also, remove them as unused, actually. :D

Done with 2eb52200c0d0f2b1a49258092d3a614d1a0d14d5 though there are lots of those cases all over the project. 😉

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
atifaziz: Sorry, this was my bad. I was only running the tests from `GeneratedTypesNamingConventionsTests` and forgot to run all the tests before doing my commit and push!

Fixed in d9a3f3e55601bdc475e5469a882a27ddfc1adfbe.

test/Amadevus.RecordGenerator.Test/RecordNamingConventionsTests.cs
atifaziz: Cheers!

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
atifaziz: > That's an important change. Could you note it additionally down in changelog?

Done in 9ea4d9cee6ed9d6550f81a38c88508f7a85eea3b.

amis92:
I'm in the middle of changing CI (appveyor->azure pipelines), so please hold on, I'll merge it when the move-over is done.

atifaziz:
> …so please hold on…

Meanwhile, wanted to say thanks for all your prompt reviews. It's fun working on a project where the maintainer is super responsive _and_ detail-oriented.

atifaziz:
Seems like you got the tests sorted out with 6430ed6d953514b5af26f8af655f39d0978a757c so I merged with master and the [tests are passing](https://dev.azure.com/amadevus/RecordGenerator/_build/results?buildId=6&view=ms.vss-test-web.build-test-results-tab) ([including on AppVeyor](https://ci.appveyor.com/project/amis92/recordgenerator/builds/26243669#L154))! 🎉

amis92:
Follow-up for #28 (analyzer asserting property names are case-insensitive different, e.g. no properties like `Count` and `count` exist for a given record type - that'd mess up the methods like Update, or constructor) in a separate PR.

PR #66, opened 2019-07-26
feat: Features flags

### LINKED ISSUE #45: Configuration in record attribute

amis92:
Implements feature flags for customizing which members are generated per-class.

Also includes support for assembly-default which, when declared, is used when no features are explicitly specified for a given class.

closes #45


amis92:
@atifaziz @jflepp please review this.

I've selected Constructor and With* for Basic aggregate feature.

The Default aggregate contains all currently implemented features. This is also, as the name suggests, the default when nothing is specified.

I'm open for discussions however in terms of those aggregates and especially what should the default be.

amis92:
For the Deconstruct, I'm pushing this as default because the [LDT design for C# Records](https://github.com/dotnet/csharplang/blob/master/proposals/records.md) states that'll be generated as well.

src/Amadevus.RecordGenerator.Attributes/RecordAttribute.cs
```diff
@@ -2,9 +2,40 @@
...
+    [CodeGenerationAttribute("Amadevus.RecordGenerator.Generators.RecordGenerator, Amadevus.RecordGenerator.Generators")]
+    [Conditional("CodeGeneration")]
+    public sealed class RecordAttribute : Attribute
+    {
+        /// <summary>
+        /// For the annotated type, generate
+        /// <see cref="DefaultRecordFeaturesAttribute.AssemblyDefaultFeatures"/>
+        /// or if not specified, <see cref="Features.Default"/>.
+        /// </summary>
+        public RecordAttribute()
```
jflepp: Would'nt it be easier to specify `Features.Default` as default value?
This way, you wouldn't have to check if a feature has been specified when calling `ConstructorArguments[0]`

src/Amadevus.RecordGenerator.Attributes/DefaultRecordFeaturesAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Amadevus.RecordGenerator
+{
+    /// <summary>
+    /// This attribute defines default <see cref="Features"/> across assembly
+    /// for usages of the non-parametrized <see cref="RecordAttribute()"/>
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = false)]
+    [Conditional("CodeGeneration")]
+    public sealed class DefaultRecordFeaturesAttribute : Attribute
```
atifaziz: You can [get away with `Features` constants in the user project](https://github.com/amis92/RecordGenerator/issues/45#issuecomment-514606896) so I would postpone adding this in a future version after some field usage. I would recommend, though, adding a note to the documentation with an example along the lines of:

```c#
static class RecordFeatures
{
    public const Features Basic = Features.Constructor
                                | Features.Withers
                                | Features.Equality;

    public const Features Internal = Basic;
    public const Features Public   = Basic | Features.Builder;
}

[Record(RecordFeatures.Public)]
partial class Foo {}

[Record(RecordFeatures.Internal)]
partial class Bar {}
```

src/Amadevus.RecordGenerator.Attributes/RecordAttribute.cs
atifaziz: This should chain to the second constructor rather than rely on property initialisation further down, as in:

```c#
public RecordAttribute() :
    this(Features.Default) {}
```

This is not a question of style but readability and centralised initialisation in the most specific constructor.

amis92: It'd not replace the check, because I'm only getting syntax-level values, and default parameter value is not taken into account here.

Also, the difference (whether it was specified or not) is what triggers the fallback to assembly-default. :)

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
```diff
@@ -7,14 +7,17 @@ namespace Amadevus.RecordGenerator.Generators
...
+        public RecordDescriptor(Features features, TypeSyntax Type, SyntaxToken TypeIdentifier, ImmutableArray<Entry> Entries, TypeDeclarationSyntax TypeDeclaration)
         {
             this.Type = Type;
             this.TypeIdentifier = TypeIdentifier;
             this.Entries = Entries;
             this.TypeDeclaration = TypeDeclaration;
+            Features = features;
         }
 
+        public Features Features { get; }
```
atifaziz: I don't think this belongs here because it feels like this is mixing description of a record with code generation options.

src/Amadevus.RecordGenerator.Attributes/RecordAttribute.cs
jflepp: Oh yea, alright that makes sense 😄

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// With* method per property that creates new record instance with
+        /// provided value assigned to the associated property. Also includes
+        /// generation of Update method.
+        /// </summary>
+        WithPerProperty = 0b_10,
+
+        /// <summary>
+        /// Basic features: <see cref="Constructor"/> and <see cref="WithPerProperty"/>.
+        /// </summary>
+        Basic = Constructor | WithPerProperty,
```
atifaziz: `Basic` could mean different things to different people so I wouldn't go there; risky business and adds little value. It pops all kinds of confusing questions like why isn't `Default` the same as `Basic` and which one is more feature-rich? It would require someone to keep looking up the definitions. Also, since `Default` _is_ more feature-rich as defined now, it gives the impression that `Basic` was needed because `Default` could be too beefy and that evokes doubt and distrust. I would just stick with `Default` as the recommended set from the project, which should frankly, give the closest implementation to [how records may get done in C#](https://github.com/dotnet/csharplang/blob/master/proposals/records.md). Again, a project can define its own `Features` constants for their record flavours/schemes/themes.

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <c>Deconstruct</c> method:
+        /// <see href="https://docs.microsoft.com/pl-pl/dotnet/csharp/deconstruct#deconstructing-user-defined-types"/>
+        /// </summary>
+        Deconstruct = 0b_10000,
+
+        /// <summary>
+        /// Default features: <see cref="Basic"/>, <see cref="ToString"/>,
+        /// <see cref="Builder"/> and <see cref="Deconstruct"/>.
+        /// </summary>
+        Default = Basic | ToString | Builder | Deconstruct,
```
atifaziz: Do we agree that this will be dynamic? If someone updates the package, they should get enhanced records if a new feature is deemed essential.

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// Record constructor that takes and assigns all read-only properties.
+        /// </summary>
+        Constructor = 0b_1,
+
+        /// <summary>
+        /// With* method per property that creates new record instance with
+        /// provided value assigned to the associated property. Also includes
+        /// generation of Update method.
+        /// </summary>
+        WithPerProperty = 0b_10,
```
atifaziz: This definitely needs a better and simpler name. I was thinking `Mutators` but that can give the wrong impression of in-place update and mutability when it's copy-and-update in actual fact. I don't have a good suggestion for now but will drop with one if it pops in my mind later.

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
atifaziz: > I don't think this belongs here because it feels like this is mixing description of a record with code generation options.

[my closing review comment](https://github.com/amis92/RecordGenerator/pull/66#pullrequestreview-267179972) says more.

src/Amadevus.RecordGenerator.Attributes/Features.cs
amis92: Yes, I'd expect it to be dynamic.

src/Amadevus.RecordGenerator.Attributes/Features.cs
amis92: At first I named it just `With`. But I've always wanted to somehow replace those separate methods with a single `With` method that'd take all parameters with default values of some generic wrapper so that those unspecified could be replaced with current values.

```csharp
public MyRecord With(
    Maybe<string> id,
    Maybe<int?> size,
    Maybe<MyRecord> inner)
{
    return Update(id.IfNone(Id), size.IfNone(Size), inner.IfNone(Inner);
}
```

That's why I didn't keep simple `With`. Maybe `WithTransforms`? I don't know.

src/Amadevus.RecordGenerator.Attributes/Features.cs
amis92: I agree. I'm not highly inclined to keep it, but still making a simple way to have ctor+withs (simpler than OR-ing it) is something that might be useful. `Barebones`? `Minimal`? Or maybe nothing indeed.

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
amis92: Let's move it to a new issue.

src/Amadevus.RecordGenerator.Attributes/RecordAttribute.cs
amis92: Good point.

src/Amadevus.RecordGenerator.Attributes/DefaultRecordFeaturesAttribute.cs
amis92: At first I wanted to defend my idea (obviously), but then I realized that the assembly-default fallback is kind-of magical, and even though I documented it, that still doesn't feel great.

There's one scenario when magic will happen anyway because if referenced Attributes package is older than Generators used, unspecified-features default will actually use the Default known to Generators and not the one in Attributes - but since it'll only happen when versions will differ, I'm not that concerned.

src/Amadevus.RecordGenerator.Attributes/Features.cs
atifaziz: > `Barebones`? `Minimal`?

Why not let a project define those?

> Or maybe nothing indeed.

Better than getting it wrong because I feel we might be over-thinking the problem. 95% of the time, `Default` should be good enough. The other 5% have a reasonable workaround (BYOF or bring your own `Features` constants). Unless you see those percentages differently?

src/Amadevus.RecordGenerator.Generators/RecordDescriptor.cs
amis92: See #71

src/Amadevus.RecordGenerator.Attributes/Features.cs
amis92: I don't see those percentages. Which convinces me to remove that. :)

src/Amadevus.RecordGenerator.Attributes/DefaultRecordFeaturesAttribute.cs
amis92: The documentation will be updated as part of #72

amis92: Also, assembly-default got dropped.

src/Amadevus.RecordGenerator.Attributes/RecordAttribute.cs
amis92: Actually, because the assembly-default behavior got dropped, there's no longer any reason to have separate constructors anyway. Merged into one with parameter with default value.

src/Amadevus.RecordGenerator.Attributes/Features.cs
amis92: Removed.

src/Amadevus.RecordGenerator.Attributes/DefaultRecordFeaturesAttribute.cs
atifaziz: > There's one scenario when magic will happen anyway because if referenced Attributes package is older than Generators used, unspecified-features default will actually use the Default known to Generators and not the one in Attributes

This is a really good point! I didn't think of that and I can't help but feel that there may be more non-obvious implications here. 🤔

src/Amadevus.RecordGenerator.Attributes/Features.cs
amis92: @jflepp @atifaziz Let's do a quick vote taking into account the above:

Name | Reaction
-|-
WithPerProperty | 👀
Withers | 🚀
With | 👍
WithTransforms | 😄
Mutators | 🎉

You can vote for multiple.

jflepp: I just like Mutators because the missing With could lead to confusion and the WithTransforms because "Transform" sounds more like a change of state to me.

atifaziz: > Let's do a quick vote

Interesting way to conduct a poll. 😄

It would have been nice to have the option to vote, “(something else)”.

amis92: You're welcome to propose other options, but I want to close this PR with *something*, probably something that's not WithPerProperty :)

atifaziz: @amis92 `CopyModifiers`? That's the best I can come up with. Perhaps I am trying too hard to come up with a _logical_ feature name but I think what kept throwing me off about `Withers` (while creative) is that it also implies `Update`. Actually, the [C# record design document](https://github.com/dotnet/csharplang/blob/master/proposals/records.md) currently [states](https://github.com/dotnet/csharplang/blob/master/proposals/records.md#with-method):

> ##### `With` method
>
> Unless there is a user-declared member named `With` declared, a record type has a compiler-provided method named `With` whose return type is the record type itself, and containing one value parameter corresponding to each *record-parameter* in the same order that these parameters appear in the record type declaration. Each parameter shall have a *caller-receiver default-argument* of the corresponding property.

This sounds awfully like `Update` (minus the defaulted arguments) so I wonder if we should align with the above and at which point, we'll have only _withers_!

jflepp: What about `WithMethods`?
Because it is being talked about a "method named `With`"

In order to be consequent with the names we could also use `BuilderClass` and `ToStringMethod()`

amis92: I think I'll go with `Withers`, since just `With` might be confusing, and `Withers` stands apart enough for a person to know what's that or check the doc comment.

atifaziz: I think that's a fair call for now. Any reason to not have _withers_ in the default set?

amis92: Haha, okay that's weird. My last comment "I think I'll go with..." was written because GitHub didn't update my page with new comments and I thought no-one responded with new proposals. Oh my oh my.

So. Appending member type to names *is* an interesting option, but it kind of feels incorrect for me because we're talking about Features, and some features like the incoming Equality will contain more members/different names than specified. `CopyModifiers` sounds too close to modifiers in syntax terms, like `public, private, internal` etc.

Essentially, the WithXyz+Update is what Roslyn implemented for their `SyntaxNode`s. It also kind of helps with creating record hierarchies, because the `Update` is then overridden in inherited class and even if you do `Base.WithA` it calls Update which resolves to `Inherited.Update` that returns new `Inherited`. That's why Update even exists, actually.

As declared, I'll merge this for now with the `Withers` feature name, but we can still continue discussing that in #45. I'm still not really satisfied with `Withers` as is. But we're still in semantic version `0` so changes like that should be easy (also changing the enum name won't break anything on the tooling side because we resolve to the int value anyway).

Issue #71, opened 2019-07-26
Reorganize and refactor internals


amis92:
> I still have my doubts about the overall approach of bit-per-feature, but I'm putting that aside for now.
> Apart from the other comments I've left, I think RecordDescriptor and generators need some refactoring because it seems that Features doesn't belong on RecordDescriptor. Unfortunately, that means some sweeping changes, and as a consequence, would take pages of comments to explain, justify & demonstrate the alternative. It would be much easier and quickly done as a PR and then use that as a basis for discussion. However, since all of those concerns are about the internal design, I've refrained from pointing them out as I am not sure what all you were looking for from this review. What I've done instead is focus on the public surface because the internals can always re-worked (let me know if you're interested in that PR).

---

>I don't think this belongs here because it feels like this is mixing description of a record with code generation options.

_Originally posted by @atifaziz in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMTg4OTc3NDE2OnYy/pull_request_review_threads/discussion_

amis92:
I'm very open for suggestions. I know this project needs a lot of refactoring and rethinking it.

For one, I'm heavily considering rewriting the whole generation to be string-based with the final step being parsing the generated syntax back to SyntaxNodes as required by CG.R framework.

It'd definitely improve readability. No idea about performance impact, but a wild guess is it wouldn't be much worse, maybe even better since we'd drop a *lot* of allocations being done when fluent-building syntax by SyntaxFactory.

amis92:
Regarding `RecordDescriptor` getting overused, I agree. We probably ought to create a record generation context that'd contain necessary information, and keep descriptor focused on syntax nodes.

amis92:
Random thought:
One of the things I think could help readability is refactoring GeneratedCodeAttributeGenerator to use SyntaxRewriter approach instead of the custom dictionary with delegates.

atifaziz:
This was addressed by 60277823d04b3725096911eb564ad79654f502ca thus close?

PR #74, opened 2019-07-31
Refactor record descriptor & generators (#71)

atifaziz:
This PR is the refactoring discussed in #71. Following is a summary of the main points.

`RecordDescriptor` no longer knows anything about `Features`; it simply describes the source record. Features influence code generation, so they are communicated to code generators via an additional parameter.

The partial generators were very inheritance-heavy and using a classic approach of boilerplate in a base class with virtual members offered so sub-classes can customize selected bits of behaviour. I refactored partial generators into a simple interface that clearly communicates what a partial generator does: it takes a record description, features and produces a _generation_:

```c#
interface IPartialGenerator
{
    Generation Generate(RecordDescriptor descriptor, Features features);
}
```

All `IPartialGenerator` implementations (`BuilderPartialGenerator`, `DeconstructPartialGenerator` & `RecordPartialGenerator`) are now singletons with their private helper methods becoming static and their dependencies becoming clear through parameters.

Since generators are singletons, they can now be safely shared and stored in an immutable array:

```c#
private static readonly ImmutableArray<IPartialGenerator> PartialGenerators =
    ImmutableArray.Create(RecordPartialGenerator.Instance,
                          BuilderPartialGenerator.Instance,
                          DeconstructPartialGenerator.Instance);
```

With the `IPartialGenerator` abstraction, you'll also notice lots of composition methods building on top of each other. For example, `RecordPartialGenerator` is a combination of _sub-generators_ that are in turn `IPartialGenerator` singletons, each implementing a feature:

```c#
internal static class RecordPartialGenerator
{
    public static IPartialGenerator Instance =>
        PartialGenerator.Combine(ConstructorGenerator,
                                 WithersGenerator,
                                 ToStringGenerator);

    private static readonly IPartialGenerator ConstructorGenerator =
        PartialGenerator.Create(Features.Constructor, descriptor =>
            Generation.Empty.AddMembers(GenerateConstructor(descriptor),
                                        GenerateValidatePartialMethod(descriptor)));

    private static readonly IPartialGenerator WithersGenerator =
        PartialGenerator.Create(Features.Withers, descriptor =>
            Generation.Empty.AddMember(GenerateUpdateMethod(descriptor))
                            .AddMembers(GenerateMutators(descriptor)));

    private static readonly IPartialGenerator ToStringGenerator =
        PartialGenerator.Member(Features.ToString, GenerateToString);

    // ...
}
```

A partial generator doesn't actually generate the whole partial type anymore. It produces a `Generation` (record) instead and which holds the base types, modifiers and members generated:

```c#
internal class Generation
{
    public SyntaxTokenList Modifiers { get; }
    public ImmutableArray<BaseTypeSyntax> BaseTypes { get; }
    public ImmutableArray<MemberDeclarationSyntax> Members;

    // ...
}
```

It is then the job of `RecordGenerator` to merge that into the actual type declaration that it creates, including common bits (which were in a base class before) like bolting on `[GeneratedCode]` attributes:

```c#
var partials =
    from g in PartialGenerators
    select g.Generate(descriptor, features) into g
    where g != null
    select
        ClassDeclaration(classDeclaration.Identifier.WithoutTrivia())
            .WithTypeParameterList(classDeclaration.TypeParameterList?.WithoutTrivia())
            .WithBaseList(g.BaseTypes.IsEmpty ? null : BaseList(SeparatedList(g.BaseTypes)))
            .WithModifiers(TokenList(g.Modifiers.Except(new[] { partialKeyword })
                                                .Append(partialKeyword)))
            .WithMembers(List(g.Members))
            .AddGeneratedCodeAttributeOnMembers();

foreach (var partial in partials)
{
    cancellationToken.ThrowIfCancellationRequested();
    generatedMembers.Add(partial);
}
```

Also note that because `partials` is lazily evaluated, cooperative cancellation is injected between each generation, which seems good enough and better than sending the token around and sprinkling the cancellation points all over. While we don't do any I/O, this is aborting on a best-effort basis if, for example, a build is cancelled.

Hope you feel this refactoring adds value by improving the code base. Let me know if you have any questions or doubts.


amis92:
In general, I quite like the approach.

Some issues I'd like to talk through:
* [Equals implementation will use `SemanticModel`](https://github.com/amis92/RecordGenerator/pull/56#issuecomment-515377759)
 - would including it in record generator be ok, or how do you see it done?
* `Generation` - the name is weird. Maybe `GenerationResult`, `GeneratorResult`?
* I'd like to, at some point, publish the generators as nuget package to allow other generators base on ours or reuse them. I'm skeptical whether making everything static is the way forward.

amis92:
I'd like to discuss my proposal of changes:
* De`static`ize methods
* Replace `Features` (and maybe Descriptor) with something like `GenerationContext` - this could include SemanticModel in future, and be extensible if we need additional objects.
* `Generation` (or whatever it's renamed to) can be extended with Diagnostics reported from generators (because equality will create some diagnostics).

atifaziz:
> In general, I quite like the approach.

Glad to hear that!

> * [Equals implementation will use `SemanticModel`](https://github.com/amis92/RecordGenerator/pull/56#issuecomment-515377759)
> 
> * would including it in record generator be ok, or how do you see it done?

I don't see a problem with either.


> * `Generation` - the name is weird. Maybe `GenerationResult`, `GeneratorResult`?

I realise we don't see _generation_ commonly used to mean the product of a generator or generation process but I saw symmetry between compiler & compilation, and generator & generation, and computer & computation, and so on. Nevertheless, I'm happy with either of your proposals if they feel more familiar.

> * I'd like to, at some point, publish the generators as nuget package to allow other generators base on ours or reuse them.

Do you think the problem domain of this project (records) is vast and variable enough to warrant or benefit from an ecosystem of generators?

> …I'm skeptical whether making everything static is the way forward.

What does static have to do with design? What's more important is if you want a design that composes well or one driven by implementation inheritance through sub-classing. The latter is often brittle and the problem of generation (input &rarr; process &rarr; output) lends itself well to functional composition and pipelines.

> * De`static`ize methods

I am not sure what you will gain. Can you give an example? May be I am not seeing this the way you are. We can also cross that bridge when the API is ready for public consumption?

> * Replace `Features` (and maybe Descriptor) with something like `GenerationContext` - this could include SemanticModel in future, and be extensible if we need additional objects.

We can add a [parameter object](https://refactoring.com/catalog/introduceParameterObject.html) that will permit the interface to remain stable while extending the context available to the generator. To some extent, the `RecordDescriptior` serves that purpose but I would not fold it with `Features` into a context object. Things that are required are best left as arguments because it makes the method very natural to read and understand through its signature.

> * `Generation` (or whatever it's renamed to) can be extended with Diagnostics reported from generators (because equality will create some diagnostics).

Make sense.

amis92:
With possible Diagnostics added, `GenerationResult` feels better for me.

I think you convinced me regarding static things. I accept this.

I'm on the fence of the parameter object. Maybe let's keep it as it is.

src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
```diff
@@ -4,46 +4,66 @@
...
+                                .WithMembers(List(g.Members))
+                                .AddGeneratedCodeAttributeOnMembers();
+
+                    foreach (var partial in partials)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        generatedMembers.Add(partial);
+                    }
+
+                    return Task.FromResult(List(generatedMembers));
```
amis92: It's a very long `case`. I'd very much prefer it to be extracted to a local function.

src/Amadevus.RecordGenerator.Generators/Generation.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Amadevus.RecordGenerator.Generators
+{
+    internal class Generation
```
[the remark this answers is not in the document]
atifaziz: While I've done this with 123a274403101898ffb6b841a9457eab61220c6e, I was wondering if `PartialGeneration` would sound better? It seems appropriate for `IPartialGenerator` to produce a `PartialGeneration` and `PartialGenerationResult` sounds mouthful (though that's what I'll rename to if you prefer).

src/Amadevus.RecordGenerator.Generators/Generation.cs
```diff
@@ -0,0 +1,56 @@
...
+        public Generation WithModifiers(SyntaxTokenList value) =>
+            new Generation(value, BaseTypes, Members);
+
+        public Generation WithBaseList(ImmutableArray<BaseTypeSyntax> value) =>
+            new Generation(Modifiers, value, Members);
+
+        public Generation WithMembers(ImmutableArray<MemberDeclarationSyntax> value) =>
+            new Generation(Modifiers, BaseTypes, value);
+
+        public Generation AddMember(MemberDeclarationSyntax member) =>
```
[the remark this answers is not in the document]
atifaziz: The `params` version is not an overload since its named different (plural). This method was just to avoid an array allocation when you're adding just one member. Happy to go with just the `params` version. In fact, I'd be even inclined to add the following:

```
public GenerationResult AddMembers(MemberDeclarationSyntax member) =>
    AddMember(member);
```

This way, the array allocation is avoided without you having to think or worry. 😉

src/Amadevus.RecordGenerator.Generators/PartialGenerator.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Linq;
+
+namespace Amadevus.RecordGenerator.Generators
+{
+    using System.Collections.Generic;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
```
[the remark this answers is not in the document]
atifaziz: Right and fixed with 9f968324fbea104f24a12633e0af067302630bc1. Sorry my default is inside the namespace declaration so this keeps slipping through.

src/Amadevus.RecordGenerator.Generators/PartialGenerator.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public static IPartialGenerator IntersectFeatures(this IPartialGenerator generator, Features features) =>
+            Create((rd, fs) => (fs & features) != Features.None
+                             ? generator.Generate(rd, fs)
+                             : null);
+
+        public static IPartialGenerator Member(Features features, Func<RecordDescriptor, MemberDeclarationSyntax> member) =>
+            Create(features, descriptor => Generation.Empty.AddMember(member(descriptor)));
+
+        public static IPartialGenerator Combine(params IPartialGenerator[] generators) =>
```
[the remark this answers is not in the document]
atifaziz: > Maybe we could make `IPartialGenerator`/`GenerationResult` ?

Not sure I get you here.

> A very random thought after reading this class. I don't know how, or from what exactly, but aggregating this, intersecting, etc. looks similar to LINQ.

Your point being? It's not surprising that it looks similar to LINQ because both use the spirit of functional composition.

src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
```diff
@@ -4,46 +4,66 @@
...
-            IEnumerable<MemberDeclarationSyntax> GenerateRecordPartials(RecordDescriptor descriptor)
+            switch (context.ProcessingNode)
             {
-                if (descriptor.Entries.IsEmpty)
+                case ClassDeclarationSyntax classDeclaration:
                 {
-                    yield break;
+                    var descriptor = classDeclaration.ToRecordDescriptor();
+                    if (descriptor.Entries.IsEmpty)
+                        goto default;
```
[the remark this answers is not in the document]
atifaziz: Hopefully 5d779207f9089f44a2bdc7d736e2b448d3233696 makes this feel less jumpy. 😉

src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
atifaziz: Okay. See 04a3efef48878e1c7d753460afc8be7bc3c7e9dc.

src/Amadevus.RecordGenerator.Generators/Generation.cs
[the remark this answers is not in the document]
amis92: Right. I'd say thought that this optimization is too early though. The allocations will happen anyway since we work on immutable nodes anyway, one allocation +/- won't impact performance (which is crazy slow anyway since we're doing a whole second compilation). However it's an additional similar method which doesn't *greatly* enhance comprehension/compression of the source code.

It can stay as is, I'm just academically discussing my POV.

src/Amadevus.RecordGenerator.Generators/PartialGenerator.cs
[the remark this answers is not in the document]
amis92: Well obviously my first sentence was not grammatically correct anyway, I think I lost some words on the brain-hand processing.

What I mean is making it LINQ-compatible via implementing IEnumerable, Select, Where, etc but again I don't really know how or why. Just an idea.

```csharp
from generator in Generators
from result in generator.Make(recordDescriptor)
group result by result.Modifiers into g
from item in g
select ...
```

It makes no sense. Let's end this conversation 😅

src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
```diff
@@ -4,46 +4,70 @@
...
+                            .WithModifiers(TokenList(g.Modifiers.Except(new[] {partialKeyword})
+                                .Append(partialKeyword)))
+                            .WithMembers(List(g.Members))
+                            .AddGeneratedCodeAttributeOnMembers();
+
+                foreach (var partial in partials)
                 {
-                    yield break;
+                    cancellationToken.ThrowIfCancellationRequested();
+                    generatedMembers.Add(partial);
```
amis92: We could do this as `yield return/IEnumerable<MemberDeclarationSyntax>`. Return type would be simpler, exiting early would be just via `yield break`, no need for the List variable. Creating the SyntaxList would need to be done on the call site, but I don't think this'd hurt readability. Thoughts?

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -76,69 +55,69 @@ StatementSyntax CreateValidateInvocation()
...
-                .WithBodyStatements(
-                    ReturnStatement(
-                        ObjectCreationExpression(
-                            Descriptor.Type)
-                        .WithArgumentList(
-                            ArgumentList(
-                                SeparatedList(arguments)))));
+            return
+                MethodDeclaration(descriptor.Type, Names.Update)
+                    .AddModifiers(SyntaxKind.PublicKeyword)
```
amis92: What formatter do you use?

I preferred the previous layout where method call's dot was on the same level as the call target.

amis92:
I'd like to wait for #56 to be merged, so that you could refactor it here as well. How do you feel about it? I fear that if this gets merged first, it'll be much harder to fix #56 merge conflicts.

src/Amadevus.RecordGenerator.Generators/Generation.cs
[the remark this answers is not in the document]
amis92: I'll leave that to you. Plain `Generation` felt too general, that's all. It's still an internal class, so what have you :)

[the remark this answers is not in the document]
atifaziz: Going the whole way with 0b693b953b9dc10fc1693a5759e4152174821ea0. 😄

src/Amadevus.RecordGenerator.Generators/PartialGenerator.cs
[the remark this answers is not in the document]
atifaziz: > What I mean is making it LINQ-compatible via implementing IEnumerable, Select, Where, etc but again I don't really know how or why. Just an idea.

This is a complex idea and it would be _academically_ interesting to see if one could express generation with monadic composition but I think the energy spent doing that wouldn't justify the end. After all, this is a project with a small scope. We should focus on _some_ internal design (this PR) and more on getting records right and going 1.0 before C# records hit the street. So, yeah, let's move along. :)

src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
atifaziz: There are a few problems with that. `yield break` means you can't test cancellation if we ever want to do that. Next, you can cancel a task but it's logically incorrect to cancel a iterable (`IEnumerable<MemberDeclarationSyntax>`). You'd have to change the return type to be `IEnumerator<MemberDeclarationSyntax>` because it's okay to cancel an _iteration_. You could still use `yield` but then then creating the final syntax list wouldn't be simple and would allocate a list per append.

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
atifaziz: I am the (human) formatter. 😆 

Sorry I missed on your styling. Do you want me to revert?

atifaziz:
I'm fine either way. I don't know how far @jflepp is from being done and I haven't been following #56 closely. While I can help @jflepp with the needed refactoring when #56 is no longer a draft, I think it might be distracting for him; let's leave it for after. What's good to know is that you're welcoming the ideas and work here and eventually intend to merge. 👍

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
amis92: I think so. Also for formatting in general, whenever I break lines within argument list, I give every argument on new line, 1 indent more than the call. I've seen a few places where arguments are both split across lines and the first is on the same as the call.

src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
amis92: That's an extremely great explanation. Thank you.

atifaziz:
FYI, 8d80250f07f2ea3c8ade96ef7355f2775a3bd3c7 fixes `PartialGenerationResult` such that any change to `PartialGenerationResult` that makes it empty will now resolve to the `PartialGenerationResult.Empty` singleton. As a result, `null` is no longer needed and shouldn't be used to mean no result, and by extension of that, it is now illegal for `IPartialGenerator.Generate` implementations to return `null`. Fewer `null` cases is a good thing.

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
atifaziz: I don't know if I got this right fcf1db2ac4cc1ab3afdeec869a424fd7e386e72e but feel free to re-format by adding to this PR.

amis92: I still see a lot of whitespace changes in the diff of this file.

atifaziz: Not sure I see your point. Can you expand?

amis92: Whitespace I mean:

image

atifaziz: Gotchya! I think this was some re-formatting done by the IDE on paste. Hopefully that noise is gone with 79db7367eff8c08ac9f296eb775b22132c512337.

amis92:
@atifaziz equality has been merged, I'll wait for conflict resolution. :)

atifaziz:
> @atifaziz equality has been merged, I'll wait for conflict resolution. :)

Unless I am mistaken, there are no unit tests for equality generators to give me the confidence that my refactoring won't break anything. 😬 Are you just looking for conflict resolution at this stage?

amis92:
Right. Umm… I guess just conflict resolution should suffice since any unit tests I'll write *shouldn't* touch the actual code.

jflepp:
Hi @atifaziz,
Im sorry about the missing tests, I will write some as soon as I have enough time.

But I would categorize the existing tests more as IntegrationTests (which are also important), in order to write UnitTests we would need more abstractions (e.g. one of the RecordDescriptor, as the semantic model is hard to fake without a mocking framework). 

What do you think?

atifaziz:
> e.g. one of the RecordDescriptor, as the semantic model is hard to fake without a mocking framework

@jflepp A way out of that would be to promote all the questions as first-class members of `RecordDescriptor` rather than hang the entire Roslyn object models off it? I'm not deep enough into the code and problem to give any sound advice here so this is a just a thought for now.

atifaziz:
@amis92 All conflicts were resolved with d989f7677422a4baefde9c1f35c3fd119da9c164 and then I made some adjustments (and fixes) with the most notable being that `PartialGenerationResult` now holds diagnostics too.

As I said earlier, I didn't follow #56 closely so I am not familiar with what was the final state of affairs. I just did a _mechanical_ refactoring to align the design. I did notice the following warnings during builds and tests:

    EqualityRecord.cs(25,33): warning CS0660: 'EqualityRecordWithOperatorEquals' defines operator == or operator != but does not override Object.Equals(object o) [A:\RecordGenerator\test\Amadevus.RecordGenerator.TestsBase\Amadevus.RecordGenerator.TestsBase.csproj]
    EqualityRecord.cs(25,33): warning CS0661: 'EqualityRecordWithOperatorEquals' defines operator == or operator != but does not override Object.GetHashCode() [A:\RecordGenerator\test\Amadevus.RecordGenerator.TestsBase\Amadevus.RecordGenerator.TestsBase.csproj]

I believe these also occur on what's in the master branch so I don't think I broke anything. I did do a diff of the code generated for types in `EqualityRecord.cs` between master and the branch of this PR and there were no changes to be seen.

amis92:
Yes, the warnings were there, definitely.

amis92:
I've added some more refactoring and cleanups myself. I feel that this is wrapped up nicely.

As part of the changes I've cleaned up the descriptors from unused parts and followed the advice of including direct answers to questions asked by generators.

So, no longer is semantic model used anywhere in descriptor data.
Descriptor constructors are clean of any logic, they are pure data objects, though still immutable.

With those changes I think writing unit tests should be easier. That's our next priority - add some unit tests to check current behavior and keep it during changes.

Issue #54, opened 2019-03-28
Validate - pass parameters by value instead of ref


amis92:
It's a (little but some anyway) overhead for reference types which are rather most of cases.

Also, Validate should probably not allow changing the actual values in constructor which it currently does.

atifaziz:
What do you think about validating post-initialization? For example:

```c#
partial class ValidatingRecord
{
    public ValidatingRecord(string name, string @switch)
    {
        this.Name = name;
        this.Switch = @switch;
        Validate();
    }

    partial void Validate();
}
```

The two downsides I see are:

1. All fields get initialised and therefore work is wasted if an argument isn't valid.
2. You cannot change the values.

I don't think point 1 is so bad because work is currently being wasted anyway pushing all arguments on the stack so is that any different then?

As for downside 2, you wanted that anyway:

> Validate should probably not allow changing the actual values in constructor which it currently does

I think this is fair but if someone wants some magic or defaulting, they should probably introduce factory methods and then making the constructor private becomes another discussion.

amis92:
Argument name is indeed lost, as you've noticed in the PR.

I'd focus on how we'd like to allow the constructor to be extended. After we agree on that we can make Validate behave accordingly.

atifaziz:
One _big_ downside of having a single & central validation method is that every with-member that updates a copy will end up validating everything, which can be wasteful for a large record!

amis92:
That's true, but we cannot have public constructor and optimize that away at the same time.

atifaziz:
> I'd focus on how we'd like to allow the constructor to be extended.

Sorry but what's the idea behind that? What kind of extensions are you thinking of?

amis92:
> I'd focus on how we'd like to allow the constructor to be extended.
> 
> Sorry but what's the idea behind that? What kind of extensions are you thinking of?

Well, since the constructor is auto-generated, the user can't add any single statement. The Validate method is at the same time an extension point that promises to always run as part of the constructor, with the same arguments.

amis92:
> One big downside of having a single & central validation method is that every with-member that updates a copy will end up validating everything, which can be wasteful for a large record!

I'd like to see how you'd solve that (decentralized validation running only on updated values).

atifaziz:
Constructor extension seems to be a different topic whereas I thought this is about validation purely. After all, the method is called `Validate` and not `OnConstructed`.

[...]

PR #83, opened 2019-08-08
💡 Make validation method post-init & parameter-less

atifaziz:
This PR demonstrates the idea discussed in [my comment](https://github.com/amis92/RecordGenerator/issues/54#issuecomment-519404007) to #54.

One side benefit of a parameter-less `Validate` method is that the syntax trees for the declaration and invocation can be shared singletons.

One new downside I discovered while taking the idea for a spin is that argument names are lost and so you have to quote and sync them manually instead of being able to rely on `nameof`. 😢 

If you like & agree with the idea, then feel free to merge the PR. Otherwise I consider it to be a demonstration/spike PR that can be closed if we conclude that this doesn't cut it.

### DIFF
--- a/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
+++ b/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
@@ -29,7 +29,7 @@ namespace Amadevus.RecordGenerator.Test
         }
 
         [Fact]
-        public void Ctor_InvokesValidate_Throwing()
+        public void With_InvokesValidate_Throwing()
         {
             var item = CreateItem();
             //
--- a/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
+++ b/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
@@ -32,7 +32,14 @@ namespace Amadevus.RecordGenerator.Test
         public void Ctor_InvokesValidate_Throwing()
         {
             var item = CreateItem();
-            Assert.Throws<ArgumentNullException>("value", () => item.WithName(null));
+            //
+            // NOTE! We test that validation is indeed invoked but don't
+            // test the parameter name since it comes from the validation
+            // method that assumes the name of the corresponding parameter on
+            // the constructor ("name") as opposed to WithName (the parameter
+            // of all with-methods is always named "value").
+            //
+            Assert.Throws<ArgumentNullException>(() => item.WithName(null));
         }
 
         [Fact]
--- a/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
+++ b/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
@@ -29,7 +29,7 @@ namespace Amadevus.RecordGenerator.Test
         }
 
         [Fact]
-        public void Ctor_InvokesValidate_Passing()
+        public void Ctor_InvokesValidate_Throwing()
         {
             var item = CreateItem();
             Assert.Throws<ArgumentNullException>("value", () => item.WithName(null));
--- a/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
+++ b/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
@@ -1,4 +1,5 @@
-﻿using Amadevus.RecordGenerator.TestsBase;
+﻿using System;
+using Amadevus.RecordGenerator.TestsBase;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Text;
@@ -25,6 +26,13 @@ namespace Amadevus.RecordGenerator.Test
             var item = CreateItem();
             var modifiedItem = item.WithName(newName);
             Assert.Equal(newName, modifiedItem.Name);
+        }
+
+        [Fact]
+        public void Ctor_InvokesValidate_Passing()
+        {
+            var item = CreateItem();
+            Assert.Throws<ArgumentNullException>("value", () => item.WithName(null));
         }
 
         [Fact]
--- a/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
+++ b/test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
@@ -29,17 +29,10 @@ namespace Amadevus.RecordGenerator.Test
         }
 
         [Fact]
-        public void With_InvokesValidate_Throwing()
+        public void With_InvokesOnConstructed_Throwing()
         {
             var item = CreateItem();
-            //
-            // NOTE! We test that validation is indeed invoked but don't
-            // test the parameter name since it comes from the validation
-            // method that assumes the name of the corresponding parameter on
-            // the constructor ("name") as opposed to WithName (the parameter
-            // of all with-methods is always named "value").
-            //
-            Assert.Throws<ArgumentNullException>(() => item.WithName(null));
+            Assert.Throws<ArgumentNullException>(nameof(Item.Name), () => item.WithName(null));
         }
 
         [Fact]
--- a/test/Amadevus.RecordGenerator.TestsBase/FeaturesTestType.cs
+++ b/test/Amadevus.RecordGenerator.TestsBase/FeaturesTestType.cs
@@ -11,9 +11,9 @@
             // and we won't see it in reflection anyway.
             // But since we have to implement it, it's already testing
             // that the generated partial declaration exists during compilation.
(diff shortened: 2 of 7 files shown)


test/Amadevus.RecordGenerator.TestsBase/FeaturesTestType.cs
```diff
@@ -11,9 +11,9 @@ public partial class FeatureConstructor
             // and we won't see it in reflection anyway.
             // But since we have to implement it, it's already testing
             // that the generated partial declaration exists during compilation.
-            static partial void Validate(ref string name)
+            partial void Validate()
             {
-                if (name is null) throw new System.ArgumentNullException(nameof(name));
+                if (Name is null) throw new System.ArgumentNullException("name");
```
amis92: I'd keep this as `nameof(Name)`. Different casing is not a huge issue here IMO, but more importantly any refactoring won't "break" this.

test/Amadevus.RecordGenerator.TestsBase/ValidatingRecord.cs
```diff
@@ -14,9 +14,9 @@ public partial class ValidatingRecord
         /// </summary>
         public string Switch { get; }
 
-        static partial void Validate(ref string name, ref string @switch)
+        partial void Validate()
         {
-            if (name is null) throw new ArgumentNullException(nameof(name));
+            if (Name is null) throw new ArgumentNullException("name");
```
amis92: I'd keep this as `nameof(Name)`. Different casing is not a huge issue here IMO, but more importantly any refactoring won't "break" this.

amis92:
In the end, I think this is the most reasonable approach to validation. After discussions in #54 I think it's the simplest of all possible approaches. It also lends itself quite well for future enhancements in other areas.

It does keep the cost of running validation on smallest changes, but the cost of avoiding it seems to be a little overblown (https://github.com/amis92/RecordGenerator/issues/54#issuecomment-519417538)

atifaziz:
> Just a nitpick to keep using `nameof` instead of literal string.

You mean this is just for this project, right? Any other project would have to hard-wire them?

I don't think we're completely out of the woods with this. Have a look at d0eff410d12e2b8797bc7d5e856194873f658f6a. The new value parameter for all `With*` members is named `value` and seeing an `ArgumentNullException` where its `ParamName` reads something else can be confusing. The demonstration in d0eff410d12e2b8797bc7d5e856194873f658f6a deliberately shows a breaking test.

In the end, I am undecided if it's a real issue or not because it's okay for methods to delegate to deeper methods that may validate and which may in turn have different parameter names.

test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
```diff
@@ -27,6 +28,13 @@ public void With_SimpleProperty_CreatesModifiedInstance()
             Assert.Equal(newName, modifiedItem.Name);
         }
 
+        [Fact]
+        public void Ctor_InvokesValidate_Passing()
+        {
+            var item = CreateItem();
+            Assert.Throws<ArgumentNullException>("value", () => item.WithName(null));
```
amis92: I'm not sure what you're trying to demo. Argument name passed to `ArgumentNullException` should reflect argument name at the place of check being done. The check is done in constructor via delegation to partial `Validate` method.

`Validate` body is written by the user, doing whatever he'll want to. And in this case I'd say that calling out Property name for the exception is as correct as it gets, because the ctor right now is just a bridge between caller and property setter. Ideally, this'd be done in the object initializer, but we can't use it 'cause we're immutable.

So to sum up, let's use nameof(Property) for our tests. In user guidelines that's also what we can advertise. And then, in the end, it's up to a developer to write the exception code.

amis92:
> In the end, I am undecided if it's a real issue or not because it's okay for methods to delegate to deeper methods that may validate and which may in turn have different parameter names.

That's exactly what I'd focus on. We're delegating checks deeper because of circumstances. The checks should use property name as that's what they're *actually* verifying, as well as because there are multiple scenarios that can throw this exception:

```csharp
[Record]
public partial class MyRecord
{
  public string Name { get; }
  partial void Validate() {
    if (Name is null) throw new ArgumentNullException("???");
  }
}

[Fact]
public void Tests()
{
  var record = new MyRecord(null); // throws, parameter name is 'name'
  record.WithName(null); // throws, parameter name is 'value'
  new MyRecord.Builder { Name = null }.ToImmutable(); // throws, no actual parameter per se
}
```

test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
```diff
@@ -32,7 +32,14 @@ public void With_SimpleProperty_CreatesModifiedInstance()
...
             var item = CreateItem();
-            Assert.Throws<ArgumentNullException>("value", () => item.WithName(null));
+            //
+            // NOTE! We test that validation is indeed invoked but don't
+            // test the parameter name since it comes from the validation
+            // method that assumes the name of the corresponding parameter on
+            // the constructor ("name") as opposed to WithName (the parameter
+            // of all with-methods is always named "value").
+            //
+            Assert.Throws<ArgumentNullException>(() => item.WithName(null));
```
amis92: Let's assume it's `nameof(Item.Name)`.

atifaziz:
> So to sum up, let's use `nameof(Property)` for our tests. In user guidelines that's also what we can advertise. And then, in the end, it's up to a developer to write the exception code.

Have a look at 4076ebde486b317355b0a1e2b2e9f935a53a4050, where validation becomes this:

```c#
        partial void Validate() =>
            Validate(Name);

        static void Validate(string name)
        {
            if (name == null) throw new ArgumentNullException(nameof(name));
        }
```

So you delegate to another method to whom you pass only the fields needing validation. The additional helper validation has the effect of then also capturing the argument names for exception via `nameof`. What do you think? I feel it strikes a nice sweet spot and the test represents guidance too. 😄

atifaziz:
I think rather than being prescriptive with a name like `Validate`, we should just call it `OnConstructed` or `OnInitialized`, which is just what it is. You can then do what you want in there, like validation. It also makes the following read better:

```c#
partial void OnInitialized() =>
    Validate(Name);

static void Validate(string name)
{
    if (name == null) throw new ArgumentNullException(nameof(name));
}
```

How do you feel about this?

test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
atifaziz: See 4076ebde486b317355b0a1e2b2e9f935a53a4050 for a compromise.

amis92:
I like the change of the name to `OnConstructed`.

Other than that, I'd call the rest as unimportant details. But for a guidance, I think it becomes over-engineered. If the validation is more complex, sure. But for the common case, I'd keep the exception throwing within the `OnConstructed` method.

As for the `nameof` usage, I think that using `nameof(Property)` is a good balance between throwing `ArgumentNullException` without argument name, and selecting arbitrary name (as is the case with the additional `ValidateName(string)`).

Property name is the thing that matters. Imagine seeing the stack trace. If the exception is thrown from within OnConstructed method with no parameters, it's a good guess that it was validating it's calling method, in this case constructor. And the constructor itself can be called by any With- method, ToImmutable, or some factory method user wrote.

---

For tests, let's use simple `OnConstructed()` with a body that throws `ArgumentNullException(nameof(Property))`.

For future guidance, do the same. Maybe suggest using custom `ValidateProperty` method to add a more meaningful stack trace entry.

test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
amis92: I'll stay with the `nameof(Property)`. It's more maintainable for our tests, and doesn't strike me as misleading anyway.

atifaziz:
> I like the change of the name to `OnConstructed`.

Done with f0e212246824379babc725a5727fe166a373ad19.

test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
atifaziz: Done with 1ff6ba45e45245343dcc73510986db2dd5f976ba.

test/Amadevus.RecordGenerator.Test/RecordMutatorTests.cs
atifaziz: Done with 1ff6ba45e45245343dcc73510986db2dd5f976ba.

PR #84, opened 2019-08-10
Fix ItemGroup vs PropertyGroup in docs

megakid:
`<DotNetCliToolReference>` must be defined in an `<ItemGroup>`, not `<PropertyGroup>` as per current docs


amis92:
Thanks!

Issue #79, opened 2019-08-06
Support struct types with a subset of features


amis92:
Providing at least a constructor, all equality implementations, ToString would be a very valuable addition.

PR #85, opened 2019-08-10
Struct support

megakid:
We have just started using this project to generate our Record types - it works nicely but we often have single property records that are best suited to be structs.  It's a fairly simple job to do this and I have also improved some of the generator code to optimize around the "single property" case.

More or less solves #79 

Please let me know if you'd like to see more of anything, it was a fairly quick job to get this far.  I couldn't work out the cleanest way to choose/override the `Features` property so ended up just declaring a `Features.StructRecommended` entry.  I have tested this with `readonly struct` types too and it all works nicely.

### DIFF
--- a/src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
+++ b/src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
@@ -133,21 +133,26 @@ namespace Amadevus.RecordGenerator.Generators
 
         private static BlockSyntax GenerateForwardToEquatableEquals(RecordDescriptor descriptor)
         {
-            // return this.Equals(obj as MyRecord);
+            // return obj is MyRecord castObj && this.Equals(castObj);
+            const string otherVariableName = "other";
             return
                 Block(
                     ReturnStatement(
+                        BinaryExpression(
+                            LogicalAndExpression,
+                            IsPatternExpression(
+                                IdentifierName(objVariableName),
+                                DeclarationPattern(
+                                    descriptor.TypeSyntax,
+                                    SingleVariableDesignation(Identifier(otherVariableName)))),
                         InvocationExpression(
                             MemberAccessExpression(
                                 SimpleMemberAccessExpression,
                                 ThisExpression(),
-                                IdentifierName(EqualsMethodName)))
-                        .AddArgumentListArguments(
-                            Argument(
-                                BinaryExpression(
-                                    AsExpression,
-                                    IdentifierName(objVariableName),
-                                    descriptor.TypeSyntax)))));
+                                IdentifierName(EqualsMethodName))
+                        ).WithArgumentList(ArgumentList(
+                            SingletonSeparatedList(
+                                Argument(IdentifierName(otherVariableName))))))));
         }
 
         private static MemberDeclarationSyntax GenerateGetHashCode(RecordDescriptor descriptor)
--- a/src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
+++ b/src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
@@ -133,7 +133,7 @@ namespace Amadevus.RecordGenerator.Generators
 
         private static BlockSyntax GenerateForwardToEquatableEquals(RecordDescriptor descriptor)
         {
-            // return obj is MyRecord castObj && this.Equals(castObj);
+            // return obj is MyRecord other && this.Equals(other);
             const string otherVariableName = "other";
             return
                 Block(
@@ -174,7 +174,7 @@ namespace Amadevus.RecordGenerator.Generators
                     PublicKeyword,
                     OverrideKeyword)
                 .AddBodyStatements(statement);
-            
+
             StatementSyntax SinglePropertyGetHashCode()
             {
                 // public override int GetHashCode() {
--- a/src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
+++ b/src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
@@ -34,19 +34,26 @@ namespace Amadevus.RecordGenerator.Generators
 
         public Task<SyntaxList<MemberDeclarationSyntax>> GenerateAsync(TransformationContext context, IProgress<Diagnostic> progress, CancellationToken cancellationToken)
         {
-            return context.ProcessingNode is ClassDeclarationSyntax cds
-                 ? GenerateAsync(cds)
-                 : EmptyResultTask;
+            // Support Class + Structs.
+            return context.ProcessingNode is TypeDeclarationSyntax tds && (tds.Kind() == SyntaxKind.ClassDeclaration || tds.Kind() == SyntaxKind.StructDeclaration)
+                ? GenerateAsync(tds)
+                : EmptyResultTask;
 
-            Task<SyntaxList<MemberDeclarationSyntax>> GenerateAsync(ClassDeclarationSyntax classDeclaration)
+            Task<SyntaxList<MemberDeclarationSyntax>> GenerateAsync(TypeDeclarationSyntax typeDeclaration)
             {
-                var descriptor = classDeclaration.ToRecordDescriptor(context.SemanticModel);
+                var descriptor = typeDeclaration.ToRecordDescriptor(context.SemanticModel);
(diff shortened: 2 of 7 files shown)


src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -133,35 +133,83 @@ private static BlockSyntax GenerateStandaloneObjectEquals(RecordDescriptor descr
 
         private static BlockSyntax GenerateForwardToEquatableEquals(RecordDescriptor descriptor)
         {
-            // return this.Equals(obj as MyRecord);
+            // return obj is MyRecord castObj && this.Equals(castObj);
```
megakid: This now works for both `struct` and `class` implementations

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -133,35 +133,83 @@ private static BlockSyntax GenerateStandaloneObjectEquals(RecordDescriptor descr
...
+
+            StatementSyntax expression;
+
+            if (descriptor.Entries.Length == 1)
+            {
+                // public override int GetHashCode() {
+                //   return EqualityComparer<TProp>.Default.GetHashCode(Prop);
+                // }
+
+                expression = ReturnStatement(HashCodeInvocation(descriptor.Entries[0]));
```
megakid: Optimization for single entry structs

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -122,20 +122,58 @@ SyntaxToken GetMutatorIdentifier()
...
-                    InvocationExpression(
-                        MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression,
-                            AnonymousObjectCreationExpression()
-                                .AddInitializers(properties.ToArray()),
-                            IdentifierName(Names.ToString))));
+            ExpressionSyntax expression;
+
+            if (descriptor.Entries.Length == 1)
+            {
+                var name = descriptor.Entries[0].Identifier;
```
megakid: Optimization for single entry structs

src/Amadevus.RecordGenerator.Attributes/Features.cs
```diff
@@ -84,6 +84,8 @@ public enum Features
         /// Equality feature set which inlcudes 
         /// <see cref="ObjectEquals"/>, <see cref="EquatableEquals"/> and <see cref="OperatorEquals"/>.
         /// </summary>
-        Equality = ObjectEquals | EquatableEquals | OperatorEquals
+        Equality = ObjectEquals | EquatableEquals | OperatorEquals,
+
+        StructRecommended = Constructor | Equality | ToString
```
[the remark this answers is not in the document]
megakid: Agreed - I didn't really like this :)

src/Amadevus.RecordGenerator.Generators/Amadevus.RecordGenerator.Generators.csproj
```diff
@@ -23,6 +23,7 @@
 
   <ItemGroup>
     <PackageReference Include="CodeGeneration.Roslyn" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" PrivateAssets="all" />
```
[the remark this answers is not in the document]
megakid: I couldn't get anything to show without red underlines in my VS2019 Pro (latest) IDE.  After some digging around, I added this and all red underlines disappeared.  Happy to take suggestions, it is a transitive dependency via `CodeGeneration.Roslyn` anyway so this doesn't actually change the output.

src/Amadevus.RecordGenerator.Generators/RecordGenerator.cs
```diff
@@ -34,20 +34,29 @@ public RecordGenerator(AttributeData attributeData)
...
-                var descriptor = classDeclaration.ToRecordDescriptor(context.SemanticModel);
+                var descriptor = typeDeclaration.ToRecordDescriptor(context.SemanticModel);
                 if (descriptor.Entries.IsEmpty)
                     return EmptyResultTask;
 
                 var generatedMembers = new List<MemberDeclarationSyntax>();
                 var features = GetFeatures();
                 var partialKeyword = Token(SyntaxKind.PartialKeyword);
 
+                var identifier = typeDeclaration.Identifier.WithoutTrivia();
```
amis92: I think this can be replaced with `descriptor.TypeIdentifier` and inlined. One variable less, reusing existing instances.

test/Amadevus.RecordGenerator.TestsBase/StructContainer.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Amadevus.RecordGenerator.TestsBase
+{
+    [Record(Features.Constructor | Features.Equality | Features.ToString)]
```
amis92: How about testing all features? (Default + Equality)

I don't think there are any issues preventing all features from working with structs.

src/Amadevus.RecordGenerator.Generators/Amadevus.RecordGenerator.Generators.csproj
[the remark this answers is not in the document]
amis92: It's hard for me to suggest any solution - "works for me". 😅

This project uses CentralPackageVersioning - and it's the only thing that comes to mind which could be problematic. But it's an Msbuildsdk, fully supported in VS2019. Sometimes VS fails with this solution, and I have to reopen it. That's about whatever troubleshooting I do. :)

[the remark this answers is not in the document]
megakid: I tried it on various other machines both here at work and at home and the dependency was always required.  Are you sure you haven't got an old version of an SDK or VS taking precedence?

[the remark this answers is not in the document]
megakid: Happy to revert and I will just keep it locally.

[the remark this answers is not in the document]
amis92: The dependency is there for sure. No reference needed, as you've yourself noted it's a transient one.

SDK is capped to v2.1.500 by `global.json` and that's the reference one. This is also the one used in cloud builds. And the cloud builds prove that this compiles as is. It's really hard for me to guess why it doesn't work for you, but if it builds in the cloud, it's correct - there's really no more neutral ground.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -133,21 +133,26 @@ private static BlockSyntax GenerateStandaloneObjectEquals(RecordDescriptor descr
...
                 Block(
                     ReturnStatement(
+                        BinaryExpression(
+                            LogicalAndExpression,
+                            IsPatternExpression(
+                                IdentifierName(objVariableName),
+                                DeclarationPattern(
+                                    descriptor.TypeSyntax,
+                                    SingleVariableDesignation(Identifier(otherVariableName)))),
                         InvocationExpression(
```
amis92: The indentation here is off. `InvocationExpression` should align with `IsPatternExpression` as an argument of `BinaryExpression`.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -133,21 +133,26 @@ private static BlockSyntax GenerateStandaloneObjectEquals(RecordDescriptor descr
...
                                 ThisExpression(),
-                                IdentifierName(EqualsMethodName)))
-                        .AddArgumentListArguments(
-                            Argument(
-                                BinaryExpression(
-                                    AsExpression,
-                                    IdentifierName(objVariableName),
-                                    descriptor.TypeSyntax)))));
+                                IdentifierName(EqualsMethodName))
+                        ).WithArgumentList(ArgumentList(
```
amis92: Use `AddArgumentListArguments` as before, and formatted appropriately.

amis92:
Thanks!

PR #86, opened 2019-08-12
Single property optimizations

megakid:


src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -122,20 +122,58 @@ SyntaxToken GetMutatorIdentifier()
...
-                            IdentifierName(Names.ToString))));
+            ExpressionSyntax expression;
+
+            if (descriptor.Entries.Length == 1)
+            {
+                var name = descriptor.Entries[0].Identifier;
+                var start = "{{ " + name + " = ";
+                var end = " }}";
+
+                expression = InterpolatedStringExpression(
```
amis92: I think this'll be much easier as a concatenation, for both generating code, and the generated code.

```csharp
public override string ToString() => "{ Prop = " + Prop + " }";
```

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -122,20 +122,58 @@ SyntaxToken GetMutatorIdentifier()
...
+            }
+            else
+            {
+                var properties =
+                    (from e in descriptor.Entries
+                     select AnonymousObjectMemberDeclarator(IdentifierName(e.Identifier)))
+                    .ToArray();
+
+                expression =
+                                        InvocationExpression(
```
amis92: Looks like some crazy indentation here :D

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
megakid: Yep, fair shout, looks much simpler in the SyntaxTree format

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -154,14 +154,14 @@ private static MemberDeclarationSyntax GenerateGetHashCode(RecordDescriptor desc
         {
             const string varKeyWord = "var";
             const string hashCodeVariableName = "hashCode";
+
```
amis92: Additional empty line not needed. There are a couple other places with additional empty lines that can be removed (within a method body).

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -154,14 +154,14 @@ private static MemberDeclarationSyntax GenerateGetHashCode(RecordDescriptor desc
...
             const int hashCodeInitialValue = 2085527896;
             const int hashCodeMultiplicationValue = 1521134295;
-            // public override int GetHashCode() {
-            //   var hashCode = 2085527896;
-            //   hashCode = hashCode * -1521134295 + EqualityComparer<TProp>.Default.GetHashCode(Prop);
-            //   ...
-            //   return hashCode;
-            // }
+
+            StatementSyntax expression = descriptor.Entries.Length == 1
```
amis92: This can be inlined, it's short.

If not, use var and change the name to `statement` or sth else, but it's definitely not an `expression` :)

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -170,40 +170,50 @@ private static MemberDeclarationSyntax GenerateGetHashCode(RecordDescriptor desc
...
-                                            Literal(hashCodeInitialValue)))))))
-                    .AddBlockStatements(
-                        descriptor.Entries.Select(EntryHashCodeRecalculation)
-                        .ToArray())
-                    .AddBlockStatements(
-                        ReturnStatement(
-                            IdentifierName(hashCodeVariableName))));
+                .AddBodyStatements(expression);
+            
+            StatementSyntax SinglePropertyGetHashCode(RecordDescriptor rd)
```
amis92: The parameter is unused, and also unnecessary.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
```diff
@@ -170,40 +170,50 @@ private static MemberDeclarationSyntax GenerateGetHashCode(RecordDescriptor desc
...
+                // public override int GetHashCode() {
+                //   return EqualityComparer<TProp>.Default.GetHashCode(Prop);
+                // }
 
-            StatementSyntax EntryHashCodeRecalculation(RecordDescriptor.Entry property)
+                return ReturnStatement(
+                        HashCodeInvocation(descriptor.Entries[0]));
+            }
+
+            StatementSyntax MultiplePropertiesGetHashCode(RecordDescriptor rd)
```
amis92: The parameter is unused, and also unnecessary.

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -122,20 +122,47 @@ SyntaxToken GetMutatorIdentifier()
...
+                : MultiplePropertiesToString(descriptor);
+
+            return MethodDeclaration(
+                        PredefinedType(Token(SyntaxKind.StringKeyword)),
+                        Names.ToString)
+                    .AddModifiers(SyntaxKind.PublicKeyword, SyntaxKind.OverrideKeyword)
+                    .WithExpressionBody(expression);
+
+
+            ExpressionSyntax SinglePropertyToString(RecordDescriptor rd)
```
amis92: The parameter is unused, and also unnecessary.

src/Amadevus.RecordGenerator.Generators/RecordPartialGenerator.cs
```diff
@@ -122,20 +122,47 @@ SyntaxToken GetMutatorIdentifier()
...
+                        LiteralExpression(
+                            SyntaxKind.StringLiteralExpression,
+                            Literal("{ " + name + " = ")),
+                        IdentifierName(name)),
+                    LiteralExpression(
+                        SyntaxKind.StringLiteralExpression,
+                        Literal(" }")));
+            }
+
+            ExpressionSyntax MultiplePropertiesToString(RecordDescriptor rd)
```
amis92: The parameter is unused, and also unnecessary.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
megakid: I prefer it with line breaks to make the code easier to read.  Write once, read many.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
megakid: It isn't recommended to use captured variables if not necessary, as you have done further up the class.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
megakid: My preference would be to not inline this sort of code as it becomes hidden in the syntax tree and therefore more difficult to read.  All of these variables will be compiled out so not sure the value in making the code so tight.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
amis92: > It isn't recommended [...]

Where?

amis92:
The styling is a matter of preference. I'd like to keep this project coherent, formatting-wise as well. Which means, please keep the style uniform, same as in other source files.

That said, I don't find a lot of line breaks helping - on the contrary, it divides lines into blocks, silently hinting about some separation. And separation means to me - maybe that's another method? So if you feel a need to separate things, refactor the block to a new method. (An exception to this rule is in unit tests, where line breaks naturally separate the three `A`s of Arrange-Act-Assert).

Style wise, in terms of formatting complicated invocation trees:
* a multiline expression starts on new line (so not in the same line as variable declaration or a return)
* every invocation has its own line
* arguments are in separate lines, 1 more indent than invocation
* fluent/dot transformations (invocations on other expressions' results) are in new line, indented the same as the target

That's how this project formats source code.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
megakid: Closures cause the compiler to generate a struct to capture the variable.  In this case it makes the code no clearer and is needless.  I am also explicitly following your own code style in the very same file.

megakid:
Understood - feel free to edit this PR to your own style.  If the style was explicitly written somewhere or enforced via `.editorconfig` you may have a better chance of keeping the formatting clean without the frustrating back-and-forth over small formatting issues.

megakid:
Wrote button

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
amis92: You prefer the code easier to read when discussing line breaks, but here you opt to decrease readability by using an obscure `rd` variable name, renaming the already available parent method's parameter for the sake of avoiding display class?

I'm not concerned with performance here. It's a code generator. It creates a completely separate, additional compilation.

amis92:
>  feel free to edit this PR to your own style
This is probably for the best then.

`.editorconfig` is on todo, however it wouldn't help much for those issues.

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
amis92: > By not using closures, I am explicitly following your own code style in the very same file.

There are no other places in this file where the local function parameter is available in parent method (that I see).

src/Amadevus.RecordGenerator.Generators/EqualityGenerators.cs
amis92: It's an interesting observation. Noted.

amis92:
Thanks!

PR #106, opened 2019-09-20
Added optional parameter withing

bboyle1234:
https://github.com/amis92/RecordGenerator/issues/102


amis92:
I'm glad you built this so quickly and easily, working simple code is the best code.

However:
* Any new Feature should be first hidden behind a new `Features` flag.
* `Attributes` namespace/package isn't the best place for embedding Optional. Also embedding is the last solution I'd like to go with. Let's keep discussing this in #102.
* Making the Optional parameters `in` doesn't really help performance, I think. It's a small struct.

